PR #3, opened 2018-10-03
design: values expression context

ericsciple:


ericsciple:
@chrisrpatterson @bryanmacfarlane @vtbassmatt

ericsciple:
@mmitche @chcosta Here is what we are thinking about the expression re-use problem. That is, the problem where you need to litter a long calculation like `${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:` many times across many files.

vtbassmatt:
This makes sense to me. Clarifying questions:

Are all `import`s resolved before `template`s? I assume yes.

When an `import` is parsed, are any directives other than `values` legal in the imported file?

ericsciple:
@vtbassmatt Yes all imports are resolved first. After all the imports are loaded, then we can process the file like normal. We might actually want to enforce that `import` comes first, since we are in reality always going to load them first before processing anything else in the file.

Within the imported file, `values` is the only thing supported currently. This gives us a place to define additional things in the future, if we want to. For example mini-templates or expression functions.

chcosta:
Is there any update on this?

AdamYoblick:
I'm very interested in when this will be available for use in VSTS as well. Having to duplicate long expressions instead of storing the results in a variable for use later is...cumbersome. :)

ericsciple:
@chcosta @AdamYoblick this isn't next on our backlog. This is similar to what we want to do, but some aspects of this make us uncomfortable about the concepts being added and customer confusion.

ericsciple:
@vtbassmatt or @chrisrpatterson can help prioritize this. Please work with them to help understand priority.

chcosta:
FYI @mmitche.  I think this is something we really want, but it's not blocking us, we can live with ugly code and documented process.  Thanks for the update @ericsciple and for helping set expectations.

It is interesting to note that when @mmitche started playing around with YAML and I started playing around with YAML and @AdamYoblick started playing around with YAML, we all unknowingly went down this same path, tried to make cleaner code, dealt with some pain, and ended up wishing for a feature like this.

AdamYoblick:
Yeah I mean even windows batch files have 1) delayed variable expansion and 2) consistent ways to access variable values at runtime.

I realize this is much more complex since there's more than just a runtime here, but an easy way to define variables to avoid copy/pasting complex expressions would certainly be nice :) Language limitations are forcing users to do things that are less than ideal.

vtbassmatt:
Thanks for pushing on this. Realistically I think this won't happen until next year since it isn't rising to the level of a blocker for y'all.

AdamYoblick:
Hey so question for @vtbassmatt and @ericsciple : Is there any way to reference ANY yaml variables in expressions, besides the environment variables like Build.Reason that we already know work? Just curious as to what is currently supported.

Thanks!

AdamYoblick:
For more context, we have a step that is included or not based on an expression that looks like this:

# If the project is NOT public AND the build is NOT for a PR, publish build assets
- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), not(contains(variables['Build.DefinitionName'], 'github'))) }}:
  - template: /eng/common-local/templates/phases/publish-build-assets.yml
    parameters:
      dependsOn:
        - Windows_NT
      queue:
        name: Hosted VS2017
      configuration: netcoreapp-Windows_NT-Debug

I'd like to change the expression to check a single variable. I'm flexible about where the variable lives (it can be a parameter to the template that runs this step, I don't care). So I just want to know if this is possible, and if it is, what the syntax should look like.

Thanks :)

vtbassmatt:
If I'm understanding the scenario right, you should be able to do something like this (done from memory, not tested)

```yaml
# template.yml
parameters:
  DO_THE_THING: false
jobs:
- ${{ if parameters['DO_THE_THING'] }}:
  - template: /eng/common-local/templates/phases/publish-build-assets.yml
    parameters: ... # all the other stuff you had
```

```yaml
# pipeline.yml
jobs:
- template: template.yml
  parameters:
    DO_THE_THING: and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), not(contains(variables['Build.DefinitionName'], 'github')))
```

AdamYoblick:
Thanks Matt, I'll give this a shot. The issue we were running into was expressions accessing global variables using the $(foo) syntax. If we can access parameters, that will work.

Trying now and will let you know. :)

AdamYoblick:
Ok so here's my question - I'm never sure what syntax to use for variables.

For example, I have this:

variables:
  # Change some variables based on build conditions
  ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest'), contains(variables['Build.DefinitionName'], 'github')) }}:
    _RunAsPublic: true
    _AgentPool: dotnet-internal-temp

Then later in the file, I have this:

# call the ci.yml template, which does the real work
phases:
- template: /eng/ci.yml
  parameters:
    runAsPublic: $(_RunAsPublic)
    agentPool: $(_AgentPool)

But $(_AgentPool) is not being expanded properly (because ci.yml is complaining that no pool exists called '$(_AgentPool)'

Do I need to do variables['_AgentPool'] instead?

vtbassmatt:
I need to get a doc up about the process of building a pipeline run from a YAML file. You're far from the only person who needs this info, and we don't have a good place for it at the moment. This comment will have the conceptual stuff; I'll address your scenario in a second comment.

There are essentially three interesting phases:
- template expansion
- orchestration planning
- agent runtime

Template expansion is very much like a Jinja or Handlebars-style templating language. The main difference is we don't operate on raw text, we operate on a YAML DOM (this mainly saves you from some of the issues related to indentation). That phase happens first, and has access to very little: just the template expression runtime functions, the DOM, and the `parameters` of the template.

After that's resolved, we do orchestration. At this phase, `$[ ]` (runtime expressions) are parsed, and we generate the graph of jobs to parcel out. Template `parameters` are gone, but `variables` are available in this phase, along with most of the built-in variables. Runtime expressions mostly result in strings today, though we're looking at how to improve this and allow richer objects as return values. I *think* we also resolve `conditions` at the job level during this phase. After runtime expressions resolve, we also expand as many `$( )` macros as we can, recursively. This may not fully expand every macro, though, and anything left is passed along to the next step.

The final phase is on the agent. The agent has an expression parser to use for step-level `conditions`, but otherwise does no expression parsing. As the agent runs steps, it gets one more shot at recursively expanding `$( )` macros. Also, the agent has additional context: the `Agent.*` series of built-in variables + any output variables passed from previous steps.

vtbassmatt:
Now to your specific scenario.

```yaml
variables:
  ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest'), contains(variables['Build.DefinitionName'], 'github')) }}:
    _RunAsPublic: true
    _AgentPool: dotnet-internal-temp

phases:
- template: /eng/ci.yml
  parameters:
    runAsPublic: $(_RunAsPublic)
    agentPool: $(_AgentPool)
```

I think what's happening is that _RunAsPublic and _AgentPool aren't ever being generated into the `variables` section. Template expansion doesn't know about variables, so that `${{ if }}` is never true. One thing I didn't mention above is that macros expand until they can't. If a macro doesn't exist, we just pass it along to the underlying task, script, etc.

You can't accomplish this at template expansion time because the data's not available until orchestration time. We've kinda come full circle, since this scenario is what the values file would help with. I think I'll have to work with you offline to figure out how to model this given the tools currently available.

AdamYoblick:
That all makes sense but the arcade yml does exactly what I'm doing (since I just copied what they had). See https://github.com/dotnet/arcade/blob/master/.vsts-ci.yml

They are using global variables instead of passing things into templates as parameters (for example, _HelixType is defined with an expression in vsts-ci.yml and referenced later at https://github.com/dotnet/arcade/blob/0d4884abeba326328accc1e994009a63dd04d4fe/eng/common/templates/phases/base.yml#L62) , which is what I was trying to do, but this doesn't seem to work with any syntax I've tried. 

One thing I didn't do was pass in the variables object which is what arcade does, but I don't understand this syntax. Do variables get passed in automatically if you declare the variables object as a parameter to a template?

chcosta:
@AdamYoblick I'm happy to walk through this with you if there's continued confusion.  Hit me up on Teams.

AdamYoblick:
Finally got this working, thanks for your help everyone. Here are the gotchas we ran into:

1. Agent pool is not able to be stored in a variable currently, because of limitations in the way VSTS parses the yaml. You CAN pass the agent pool as a parameter, but at some point, it must be "compile time constant"

2. The runAsPublic input that was added to all the arcade yaml templates also must be hardcoded as some point. This is because template expansion ( ${{ if ... }} ) has no access to user defined variables, and this input is indeed used in template expansion later on.

ericsciple:
I'm closing this PR. Instead here is what we are [planning](https://github.com/Microsoft/azure-pipelines-yaml/pull/46)

The new PR solves some of the same problems, without adding new concepts. Please review the new PR

PR #8, opened 2018-10-10
design: container steps

vtbassmatt:


design/container-steps.md
```diff
@@ -0,0 +1,55 @@
...
+```yaml
+resources:
+  containers:
+  - image: microsoft/dotnet:latest
+    name: dotnet
+  - image: facebook/yarn:latest
+    name: yarn
+
+steps:
+- run: dotnet
```
chrisrpatterson: In the case of a container that is not on a private registry you should be able to
```
- run: microsoft/dotnet:latest
```
we can simply assume if we do not find a container resource with the matching name that this must be an image and we can simply run it.

vtbassmatt: @ericsciple and I talked about that. I wanted to be cautious about overloading the meaning of `run: <string>` with too much magic. If I have a string `- run: python`, that'll pull `python:latest` right up until the day I add a resource called `python`.

We're adding inline resource parsing, though, so we can do:
```yaml
- run:
    image: microsoft/dotnet:latest
```

Let me think about it some more, though - I really like the super abbreviated form.

chrisrpatterson: Sure it would change the meeting of -run: python but it would be the result of an explicit action you took.  Given we have no compatibility issues here I think we can push the limits a bit compared to other areas where compatibility with existing features forced a slightly more verbose syntax.

vtbassmatt: Yeah, works for me. I'll update.

vtbassmatt:
Merging so that we have a locked version to use as a spec.

PR #12, opened 2018-10-16
spec: scheduled build syntax

vtbassmatt:


design/schedules.md
```diff
@@ -0,0 +1,76 @@
...
+    - releases/*
+  batch: true
+  always: true
+```
+
+These definitions say:
+- For `master` and all `releases/*` branches except those under `releases/ancient/*`, run a pipeline at midnight every day.
+- For all `releases/*` branches, run a pipeline at noon on Sundays.
+  - Only run against branches which have changed (the `always` keyword).
+  - Only run one instance of the pipeline per branch (the `batch` keyword). If a branch already has an active run, queue the pipeline instead of running it immediately.
```
ericsciple: do we need `batch`? aren't queues the throttling mechanism? Either demand a specific agent or create a separate queue for throttling.

design/schedules.md
```diff
@@ -0,0 +1,76 @@
...
+    include:
+    - releases/*
+  batch: true
+  always: true
+```
+
+These definitions say:
+- For `master` and all `releases/*` branches except those under `releases/ancient/*`, run a pipeline at midnight every day.
+- For all `releases/*` branches, run a pipeline at noon on Sundays.
+  - Only run against branches which have changed (the `always` keyword).
```
ericsciple: > Only run against branches which have changed

How do you expression that? `always: false` ? What is the default?

design/schedules.md
```diff
@@ -0,0 +1,76 @@
...
+Accepted values are:
+
+Field | Range
+------|------
+Minutes | 0 through 59
+Hours | 0 though 23
+Days | 1 through 31
+Months | 1 through 12, full English names, first three letters of English names
+Day of week | 0 through 6, full English names, first three letters of English names
+
```
ericsciple: other calendars? future maybe?

vtbassmatt: Good idea for the future. We could accept other input formats beyond just `- cron:`.

design/schedules.md
vtbassmatt: We need the behavior of "I want to only run one of these at a time; queue if there's one already running". Much like `batch: true` on `ci`, right?

design/schedules.md
vtbassmatt: I messed up the human-readable string here :) It should say "Run even against branches which haven't changed".

Issue #17, opened 2018-10-17
add template for R


bstabler:
I've put together a template for R that I think others would be interested in.  I'll create a pull request for you to review.

PR #18, opened 2018-10-17
add simple R example

bstabler:
#17

### DIFF
--- a/templates/R.yml
+++ b/templates/R.yml
@@ -1,0 +1,22 @@
+# R
+
+pool:
+  vmImage: 'ubuntu-16.04'
+
+steps:
+- script: sudo apt-get update
+  displayName: 'Update packages'
+
+- script: sudo apt-get install r-base
+  displayName: 'Install R'
+
+- script: |
+    cd /usr/local/lib/R
+    sudo chmod o+w site-library
+  displayName: 'Change permissions for package installs'
+
+- script: Rscript -e 'install.packages("devtools", repos = "http://cran.us.r-project.org")'
+  displayName: 'Install devtools'
+
+- script: Rscript -e 'print("hi")'
+  displayName: 'Run R expression'


chrisrpatterson:
Wouldn't it make more sense to rely on the r docker container here instead of installing?

https://hub.docker.com/_/r-base/
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=vsts&tabs=yaml

vtbassmatt:
@bstabler given your +1 to Chris's comment, are you going to switch to a container approach? If not, I'll close this for now.

bstabler:
I think both examples are useful.  How about I (or someone else) add a docker version as well?

vtbassmatt:
The templates are intended to show the canonical / most common way, not to provide options. A pattern we've used before is to keep the template simple but point to a language guide in our docs (for example, [Python](https://docs.microsoft.com/azure/devops/pipelines/languages/python)) that goes into more detail. We take doc PRs in [another repo](https://github.com/MicrosoftDocs/vsts-docs/) if that's a line you want to pursue.

bstabler:
I updated the example to use the R docker container, thanks.

lorenzwalthert:
Why have you removed the R package installation example? I think it could be pretty informative and also, everyone will want to install packages. I am struggling at the moment to get R package installed in the pipeline. I checked https://github.com/rocker-org/rocker-versioned/issues/108 and then tried to make the folder 777, but I did not have sufficient permission.

nickforr:
Hi,

If you're looking for an example for installing packages, I'd suggest adopting the same approach as is in the rocker images (using install2.r from Dirk's littler library). For example, from the [tidyverse](https://github.com/rocker-org/rocker-versioned/blob/master/tidyverse/Dockerfile) container in rocker:

```
install2.r --error \
    --deps TRUE \
    tidyverse \
    dplyr \
    devtools \
    formatR \
    remotes \
    selectr \
    caTools \
    BiocManager
```

vtbassmatt:
Hey all. We don't have an R expert or champ internal to the team. Does the PR as it currently exists look good? If a couple of you want to give me a thumbs-up, I'll be happy to merge it. Otherwise, I'm going to close this.

PR #19, opened 2018-10-17
Enable publishing of JUnit results by default

mikeharder:
I believe most of the pipeline templates publish test results by default, so I think the Java templates should do the same, unless there's a good reason not to.

### DIFF
--- a/templates/android.yml
+++ b/templates/android.yml
@@ -12,6 +12,6 @@ steps:
     workingDirectory: ''
     gradleWrapperFile: 'gradlew'
     gradleOptions: '-Xmx3072m'
-    publishJUnitResults: false
+    publishJUnitResults: true
     testResultsFiles: '**/TEST-*.xml'
     tasks: 'assembleDebug'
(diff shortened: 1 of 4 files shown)


DavidStaheli:
Before we merge this...
We currently issue a build warning if test result publishing is enabled but no tests are found ([example](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/GradleV2/gradletask.ts#L46)).  This will cause new pipelines to get ugly warnings if they have no test results.  😢

I'd love to change that warning to informational output.  But that changes behavior, such as if a task runs conditionally based upon another task having no warnings.  So that probably warrants a new major version.  Do you agree?

DavidStaheli:
Yes but it seems like the default task settings shouldn't result in a build with warnings.  Builds will be dirty for people just starting out.

chrisrpatterson:
There is no way to conditionalize a task based on another task having warnings so that is not an issue as these warnings do not indicate partial success for the task.  That being said I agree that we should not have warnings for missing files.  However, we do need some way to indicate to the user that the pattern supplied does not match any files in the case that they did intend to publish test results but mistyped the patttern.

leantk:
Closing this because it is old. @mikeharder if you are still looking to get this in, please re-open it and we will go over the pros/cons that were hashed out above

PR #41, opened 2018-11-01
Pipeline variables

vtbassmatt:


alexmullans:
Nice! We've definitely hit the Build.* vs. Release.* issue with the packaging-related tasks, so I'm excited to get to a common variable future :).

vtbassmatt:
@TingluoHuang

vtbassmatt:
Not yet mentioned: new "Pipeline.ArtifactsDirectory", "Pipeline.DefaultSourcesDirectory", and a few others in my head. TODO.

vtbassmatt:
Not mentioned yet: putting some values into expression context and letting customers opt into them.

design/variables-in-pipelines.md
```diff
@@ -0,0 +1,106 @@
...
+| _TODO_
+
+Necessary new variables:
+
+| New variable | Description | Special notes |
+|--------------|-------------|---------------|
+| CI | Set to "True" to match industry expectation for CI systems | Environment only, not available in expressions
+| AZURE_PIPELINES | Set to "True" to differentiate from other CI systems | Environment only, not available in expressions
+| Pipeline.Workspace | Root directory where all source, artifacts, etc will be placed
+| Pipeline.Url | https:// URL to pipeline definition | [requested](https://twitter.com/_a__w_/status/1102802095474827264)
```
chrisrpatterson: I think the link that was requested was the URL of the running pipeline and specific the logs page, not the definition.

vtbassmatt: 🤦‍♂️  that's embarrassing

design/variables-in-pipelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+| Current variable | Keep, cut, or rename |
+|------------------|----------------------|
+| _TODO_
+
+Necessary new variables:
+
+| New variable | Description | Special notes |
+|--------------|-------------|---------------|
+| CI | Set to "True" to match industry expectation for CI systems | Environment only, not available in expressions
```
willsmythe: Case of the value will matter (`True and `true`). I think we should aim to be consistent with other CI services, which tend to use lowercase strings for booleans.

vtbassmatt: good catch, thx

PR #42, opened 2018-11-01
Pipeline Artifacts proposal

alexmullans:
**(Note, below is the original PR text. For the best experience, start with "Files changed", which has the final proposal.)**

Moving this over from a private repo. Also separating out this proposal from the use/auth proposal (#40).

@ashokirla can you take a look at this and see if I've covered all your feedback? I looked at your Word doc and tried to hew as closely as possible to your `downloadArtifacts` section, but I couldn't do it 100% since you can't have a value and then a list (e.g. [1]). So, I instead propose providing the resource as a separate key (see line 43 of this file). Thoughts?

I think we're pretty close to locking this in, which means we can then schedule the work:
1. On Artifacts Core, ensuring the tasks support the options we want to expose here, and updating the tasks so that the key names match what we're proposing here (so that whether you use the task or the shortcut, you provide the same keys).
2. On CI Platform, implementing the `artifact` and `downloadArtifacts` shortcuts that call out to the tasks in (1.)

Anything I'm missing? Do we need any more review other than approving this PR and merging it into the design repo?

[1]
```yaml
  - downloadArtifacts: self
    artifactName: WebTier1
```

### DIFF
--- a/design/pipeline-artifacts.md
+++ b/design/pipeline-artifacts.md
@@ -22,7 +22,7 @@ In this document, `artifact` and `downloadArtifact` refer specifically to Pipeli
 ### Example
 
 ```yaml
-- publishArtifact:
+- artifact:
   - **/bin/*
   - **/obj/*
   name: webapp
@@ -38,18 +38,16 @@ In this document, `artifact` and `downloadArtifact` refer specifically to Pipeli
 ### Schema
 
 ```yaml
-- downloadArtifacts:
-  - name: string # identifier for the artifact to download, optional; defaults to 'default'
-    patterns: string | [ string ] # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to download; if blank, the entire artifact is downloaded
-    root: string # the directory in which to download files, defaults to $PIPELINES_RESOURCESDIR
-    resource: # the resource from which to download the artifact, defaults to self (indicating the current pipeline)
+- downloadArtifact: # identifier for the artifact to download, optional; defaults to 'default'
+  patterns: string | [ string ] # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to download; if blank, the entire artifact is downloaded
+  root: string # the directory in which to download files, defaults to $PIPELINES_RESOURCESDIR
+  resource: # the resource from which to download the artifact, defaults to self (indicating the current pipeline)
 ```
 
 ### Example
 
 ```yaml
-- downloadArtifacts:
-  - name: webapp
+- downloadArtifact: webapp
 ```
 
 ---
@@ -92,9 +90,8 @@ You can control the location where artifacts are downloaded using the `downloadR
   - artifact: bin/*
 - job: Deploy
   steps:
-  - downloadArtifacts:
-    - name:
-      downloadRoot: $(Pipelines.SourcesDir)/from-build/
+  - downloadArtifact:
+    downloadRoot: $(Pipelines.SourcesDir)/from-build/
   - script: |
       TODO-some-cool-deploy-script-here $(Pipelines.SourcesDir)/from-build/bin/
 ```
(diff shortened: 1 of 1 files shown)


design/pipeline-artifacts.md
```diff
@@ -0,0 +1,145 @@
...
+- artifact: string | [ string ]  # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to publish
+  name: string # identifier for this artifact (no spaces allowed), defaults to 'default'
+  prependPath: string # a directory path that will be prepended to all published files
+  seal: boolean # if true, finalizes the artifact so no more files can be added after this step
+```
+
+### Example
+
+```yaml
+- publishArtifact:
```
vtbassmatt: This should just be `- artifact`

design/pipeline-artifacts.md
```diff
@@ -0,0 +1,145 @@
...
+---
+
+## Proposal: downloadArtifact
+
+`downloadArtifact` is a shortcut for the [Download Pipeline Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact.md) task. It will download artifacts published from a previous job or stage or from another pipeline. Artifacts are downloaded either to `$PIPELINES_RESOURCESDIR` or to the directory specified in `root`.
+
+### Schema
+
+```yaml
+- downloadArtifacts:
```
vtbassmatt: I want to flip this to match the `- artifact` pattern. Schema:
```yaml
- downloadArtifact: string # artifact name
  patterns: string | [ string ]
  ...
```
and you can say this multiple times as "steps".

design/pipeline-artifacts.md
```diff
@@ -0,0 +1,145 @@
+# Pipeline Artifacts YAML shortcut
+
+**Status: PM spec**
+
+Pipeline Artifacts are the new way to move files between jobs and stages in your pipeline. They are hosted in Azure Artifacts and will eventually entirely replace FCS "Build Artifacts". Because moving files between jobs and stages is a crucial part of most CI/CD workflows, and because Pipeline Artifacts are expected to be the default way to do so, this spec proposes a YAML shortcut syntax for publishing and downloading artifacts.
+
+In this document, `artifact` and `downloadArtifact` refer specifically to Pipeline Artifacts. Artifacts are distinct from `resources`, which are various kinds of inputs like `pipelines`, `containers`, `repositories`, and `packages`. Ashok is driving the YAML spec for these inputs, and he and I (amullans) are in sync.
```
vtbassmatt: I like this. I'm open to (and would like) to offer a provider model syntax for different kinds of things, but I'd also like to keep Pipeline Artifacts pure and first-class.

alexmullans:
@vtbassmatt I reacted to your feedback, all good. I'll have a final chat with @ashokirla when I'm in India next week about the "pipeline artifacts vs other build systems vs other artifact types" scoping question, then I think we can close this out and talk about scheduling the week of Thanksgiving.

design/pipeline-artifacts.md
```diff
@@ -35,22 +35,53 @@ In this document, `artifact` and `downloadArtifact` refer specifically to Pipeli
...
+```
+
+### Option 3: multi-artifact in one step
+
+#### Schema
+
+```yaml
+- downloadArtifacts: string # identifier for the pipeline from which to download artifacts, optional; defaults to 'self'
+  version: string # the run/version of the pipeline to get artifacts from
+  - artifact: string # identifier for the artifact to download
```
vtbassmatt: Ignoring that this isn't valid YAML, I also don't want to support two ways to do the same thing.

design/pipeline-artifacts.md
```diff
@@ -35,22 +35,53 @@ In this document, `artifact` and `downloadArtifact` refer specifically to Pipeli
...
 
 ```yaml
 - downloadArtifact: webapp
+- downloadPipeline: P2
+  artifacts: p1
+  version: 2
 ```
 
----
+### Option 2: fewer keywords
```
vtbassmatt: I much prefer this. `downloadPipeline` doesn't read right to me.

PR #46, opened 2018-11-07
design: variables context and condition simplification

ericsciple:


ericsciple:
offline review:
- lazy evaluate variable expressions? future.
- parseBool? kill this for now
- else? elif too (handlebars does `else if`)
- templates: allow parameters
- don't flow downstream across files
- stomp over if already defined

ericsciple:
@mmitche @chcosta @MichaelSimons @AdamYoblick let us know what you think.

We're tabling the "values" concept for now. We're trying to solve some of the same problems, but without introducing new concepts.

design/variables.md
```diff
@@ -0,0 +1,71 @@
...
+Example jobs template:
+
+```yaml
+parameters:
+  jobName: ''
+jobs:
+- job: ${{ parameters.jobName }}
+  variables:
+    public: ${{ eq(variables['System.TeamProject'], 'public') }}
+    publicOrPR: ${{ or(eq(variables.public, 'true'), eq(variables['Build.Reason'], 'PullRequest')) }}
```
chcosta: I imagine that switching between access methods will cause confusion, ie, people will want to either always do a hash table lookup method (`variables['key']`) or an object property method (`variables.key`).  Asking people to understand the context and switch between the methods may lead to frustration.

design/variables.md
```diff
@@ -0,0 +1,71 @@
...
+
+```yaml
+# my-jobs.yml
+
+parameters:
+  jobName: ''
+jobs:
+- job: ${{ parameters.jobName }}
+  variables:
+  - template: my-variables.yml # reference variables
```
chcosta: I presume that this is valid?

```
variables:
- template: my-variables.yml
  parameters:
    config: release
- name: myvar
  value: foo
```

ie, you can define the variables object via a combination of one (or more) templates and some explicit variable declarations?

MichaelSimons:
> We're tabling the "values" concept for now. We're trying to solve some of the same problems, but without introducing new concepts.

@ericsciple - Do you have a timeline for solving the "reusable pools" problem?  Is adding support for yaml anchors something you are considering?  I know it isn't the ideal solution but it helps in some cases.

design/variables.md
ericsciple: yes

design/variables.md
ericsciple: they are interchangeable, but the dot notation is restricted (member name must consist of a-z0-9_)

ericsciple: ...and that is current behavior already. Nothing is changing there.

AdamYoblick: Eric, I actually ran into a bug yesterday which proved that they are not interchangeable. Only the hash table syntax works every time, the dot notation does not. Surprised me :)

chcosta: Do you have a sample of the bug?  My understanding (based on ad-hoc testing) is that they are interchangeable modulo the caveat which Eric outlined above.

AdamYoblick: Yep, use any of the template expressions in your yaml:

${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), not(contains(variables['Build.DefinitionName'], 'github'))) }}:

Replace all of these with the dot notation and it doesn't work as it should. Switch it back to hash table notation and it does.

AdamYoblick: At least that was my experience. I much prefer the dot notation because it's less clutter, less typing, and more in line with json and normal property lookups. So I tried switching them but it didn't work properly.

chcosta: > member name must consist of a-z0-9_

`.` is treated differently in "dot" notation.

ericsciple:
Else and elif has not landed yet, but adding variables to the context and variables templates is in master. It will ship with our sprint 148.

chcosta:
Somehow I missed that this had merged.  I've validated that this is available in our organization, I'm very excited to cleanup our templates thanks to this change.

FYI @mmitche @jcagme @riarenas @tmat @AdamYoblick

PR #52, opened 2018-11-14
port over sidecar spec from private repo

vtbassmatt:


vtbassmatt:
@dakale here's the spec. I'm going to merge the PR but feel free to comment.

PR #53, opened 2018-11-15
additions to the Use spec

vtbassmatt:


vtbassmatt:
- `authWith`?
- `use`?
- share out more details on auth syntax

vtbassmatt:
- maybe only support auth to one, since it simplifies the designer

PR #54, opened 2018-11-16
spec: resources in YAML

ashokirla:
This is the spec for pipelines, repositories, containers as resources.


ashokirla:
Refer to Pipeline-artifacts proposal: https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md

design/pipeline-resources.md
```diff
@@ -0,0 +1,190 @@
...
+    type: azurePipelines
+    source: 
+      name: DevOpsProject/SmartHotel-CI  # name of the pipeline source from different project
+      branch: releases/M142
+```
+
+### `downloadArtifact` for pipelines
+
+Artifacts from the `pipeline` resource are automatically downloaded and made available for all the jobs. However, in any of the jobs, you can choose to override and download only specific artifacts using `downloadArtifact` shortcut.
+
```
vtbassmatt: Mention something about "if you override a download, all automatic download behavior is removed. You're responsible for specifying everything you need".

design/pipeline-resources.md
```diff
@@ -0,0 +1,190 @@
...
+## Resources: `repositories`
+
+If you have multiple repositories from which you need to sync the code into your pipeline, you can consume the repos by defining a `repositories` resource. A repository can be another Azure Repo or any external repo like GitHub etc.
+
+
+### Schema
+
+```yaml
+resources:          # types: pipelines | repositories | containers | packages
```
vtbassmatt: This section needs changes. The resource doesn't need all the imperative stuff around how much to fetch and whether to checkout LFS files. That belongs in `checkout`.

See [docs on what's in the repositories resource](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=vsts&tabs=schema#repository) and [what's in the checkout step](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=vsts&tabs=schema#checkout). I also have a [spec up for multiple checkout](https://github.com/Microsoft/azure-pipelines-yaml/pull/60). If it needs amendments, feel free to comment on the PR.

design/pipeline-resources.md
```diff
@@ -0,0 +1,190 @@
...
+```yaml
+resources:         
+  containers:
+  - container: devLinux      
+    image: GitHub
+    connection: myDockerRegistry
+    source: Microsoft/alphaworz
+```
+
+Container image and metadata from the resource are made available across all the jobs in the pipeline for cosumption.
```
vtbassmatt: Can you clarify what this means? I think it's things like making variables available for `condition`s and the like.

vtbassmatt:
@ericsciple @patrickcarnahan @bryanmacfarlane @chrisrpatterson

design/pipeline-resources.md
ashokirla: Yeah, now that we are enabling multiple repo checkout support, there are few scenarios we are missing out with your proposal here: https://github.com/Microsoft/azure-pipelines-yaml/blob/3a778fa86fa2536ff64b97bfdb17a13ba4e23d08/design/multicheckout.md

1. User should be aware of the defaults that are taken when a repo is checked out by default. For ex. What is the fetchDepth, are the creds persisted etc. 

2. If user choose the change the default behavior of the checkout, then he/she have to define checkout explicitly across all the jobs for all the repositories. This will be tedious and bloats up the YAML file. 

Moreover, these are optional fields and user doesn't have to provide if not required. And user can always override within the checkout step. By providing support for these details within the resource, the process can get simplified. Please share your thoughts on it.

design/pipeline-resources.md
ashokirla: Sure, updated the info appropriately. Btw, we are currently working on the proposal for ACR and Helm specs. I will share the detailed proposal on containers along with triggers as a separate PR.

design/pipeline-resources.md
vtbassmatt: I'm still not convinced. Having more than one "right" way to do things (defaults from the resource vs. overrides on the checkout step) will lead to confusion. I expect templates to address the repetition problem; if we're missing a feature there, let's fix that.

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+When you use `checkout` to sync a specific repository resource, all the other repositories are not synced. 
+
+`self` checkout directory: `$PIPELINES_SOURCESDIR`
+other repositories' checkout directory: `$PIPELINES_RESOURCESDIR/<repository-identifier>/`
+
+## Resources: `containers`
+
+If you need to consume a container image as part of your CI/CD pipeline, you can achieve it using `containers`. A container can be an Azure Container Registy or any external Docker registry.
```
alexmullans: "Regist**r**y" (missing r)

alexmullans: Can you clarify that this is specifically about **running one or more jobs inside a container** and not about building containers?

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+If you have a pipeline that produces artifacts, you can consume the artifacts by defining a `pipelines` resource. A pipeline can be another Azure DevOps pipeline or any external pipelines like Jenkins etc.
+
+
+### Schema
+
+```yaml
+resources:        # types: pipelines | repositories | containers | packages
+  pipelines:
+  - pipeline: string  # identifier for the pipeline resource
+    type: enum  # type of the pipeline source like azurePipelines, Jenkins etc.
```
alexmullans: @vtbassmatt is "azurePipelines" the standard keyword/type here? Also, @ashokirla, could we have `azurePipelines` be the default type, so that in the common case you don't have to provide `type` at all?

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+### Examples
+
+If you need to consume another azurePipelines from the current project and you dont require setting branch, version and tags info, this can be shortened to:
+
+```yaml
+resources:         
+  pipelines:
+  - pipeline: SmartHotel      
+    type: azurePipelines
+    source: SmartHotel-CI  # name of the pipeline source definition
```
alexmullans: Is this valid YAML? I don't think so but @vtbassmatt is a better judge than I am.

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Schema
+
+```yaml
+resources:        # types: pipelines | repositories | containers | packages
+  pipelines:
+  - pipeline: string  # identifier for the pipeline resource
+    type: enum  # type of the pipeline source like azurePipelines, Jenkins etc. 
+    connection: string  # service connection to connect to the source
+    source:
```
alexmullans: I get why `source` needs to be its own dictionary now, since Jenkins and Azure Pipelines (and TeamCity, etc.) will probably want different inputs, but it seems a bit kludgy. Could we maybe elevate some of what's in `source` to be top-level keys? Then, maybe use something like `inputs: { string: string }` for additional, source-specific information, so that we're consistent with Tasks? @vtbassmatt what do you think?

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+Or to avoid downloading any of the artifacts at all:
+
+```yaml
+- downloadArtifact: none
+```
+
+
+Refer to [download artifacts](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md#downloading-artifacts-downloadartifact) for more details.
+
+Artifacts from the `pipeline` resource are downloaded to `$PIPELINES_RESOURCESDIR/<pipeline-identifier>/` folder.
```
alexmullans: Can you clarify this? Isn't it something more like "Artifacts from the `pipeline` resource are downloaded into named folders in `$PIPELINES_RESOURCESDIR/<pipeline-identifier>`, e.g. `$PIPELINES_RESOURCESDIR/SmartHotel/default`?

design/pipeline-resources.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Resources: `containers`
+
+If you need to consume a container image as part of your CI/CD pipeline, you can achieve it using `containers`. A container can be an Azure Container Registy or any external Docker registry.
+
+### Schema
+
+```yaml
+resources:          # types: pipelines | repositories | containers | packages
+  containers:
```
alexmullans: How does this match with what @vtbassmatt already has in the [Containers section](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=vsts&tabs=schema#container)? Are we changing `endpoint` to `connection`?

design/pipeline-resources.md
chrisrpatterson: These are properties of explicitly how to check out the repo and not about the repo resource itself.  Those properties belong with the checkout step.

design/pipeline-resources.md
ashokirla: yes, it is a valid YAML. Please refer to the image below.

yaml syntax

design/pipeline-resources.md
ashokirla: yes, suppose smartHotel pipeline publishes two artifacts 'Web' and 'DB'. There will be two artifacts downloaded to $PIPELINES_RESOURCESDIR/SmartHotel/ i.e. `$PIPELINES_RESOURCESDIR/SmartHotel/Web` and `$PIPELINES_RESOURCESDIR/SmartHotel/DB`.

design/pipeline-resources.md
ashokirla: Thanks again :).

design/pipeline-resources.md
ashokirla: yeah, today service endpoints are renamed to service connections. And wherever possible we are using the word `connection` instead of `endpoint`. I think we should change this.

design/pipeline-resources.md
ashokirla: Okay, makes sense. I will remove the repo properties from resources and will find out if templates can solve this repetition problem.

design/pipeline-resources.md
ashokirla: We looked at the inputs approach as well. Ultimately, both are similar approaches. This approach increases readability. I leave it to @vtbassmatt to suggest the better approach.

design/pipeline-resources.md
ashokirla: Yeah, when a container image is defined as part of resources, all the container specific info like image, connection, tags and other meta-data is passed to all the jobs in the pipeline. This info can be consumed as part of your K8s deployment to extract the image pull secrets and provide it to K8s cluster to download the image and deploy to the cluster.

design/pipeline-resources.md
ashokirla: Discussed with @vtbassmatt and @chrisrpatterson about the schema. The inputs starting from `project` to `tags` are provider `type` specific properties. The above schema is for the type: azurePipelines. The inputs may change for other provider types like Jenkins etc.

design/pipeline-resources.md
ashokirla: Agreed by @vtbassmatt. For backward compatibility we will still support `endpoint` as a deprecated input. But going forward we will use `connection`.

PR #57, opened 2018-11-19
Proposals/pipeline caching

mitchdenny:
Starting the review process for a proposal to add caching to Azure Pipelines. This is a fairly large proposal and needs a bit of time to digest. The intention is to cover off as many scenarios as possible and then if we decide to go ahead prioritize what features we deliver first since there is a lot here.


design/pipeline-caching.md
```diff
@@ -0,0 +1,366 @@
...
+
+### The basics & task integration
+
+When work is done on an agent as part of a pipeline, it is represented as a _job_ when a set of _steps_ which are implemented by tasks. For many pipelines the YAML syntax obscures the underlying "tasky" nature of Azure Pipelines, but they are there nonetheless. As a result I'm starting with how pipeline caching will integrate with the underlying task infrastructure.
+
+Detecting a cache hit or miss will be performend by the agent worker. As the worker prepares to execute each task it will check to see whether any given task has caching enabled and if so it will determine the key and check for a cache hit. If a hit is detected it will set about restoring the cached content into the pipeline execution environment. Exactly how that content is restored will be discussed a bit later.
+
+The benefit of this approach is that caching can be applied to all existing tasks without any additional work by the task author. Support for caching can be enhanced to provide sensible defaults (e.g. what files to consider for cache key input).
```
chrisrpatterson: I don't think this is going to be correct in many cases.  For the most part you will likely still want to run the task to normalize the cache to make sure you have the right versions as there are a lot of cases where you might choose a key like branchname and that would give you mostly the right files and then you only have to download the delta from the package source.

design/pipeline-caching.md
```diff
@@ -0,0 +1,366 @@
...
+
+Provided the cost of determining a cache hit and acquiring the contents of the cache is cheaper than the cost of producing the output again from scratch then the pipeline's performance will be increased.
+
+Caching can introduce many complexities into build pipelines and our goal is to strike the balance between "correct" and "good enough" where "good enough" allows us to provide performance benefits without a disproporionate amount of effort on behalf of the user.
+
+## Proposal
+
+This proposal contains a number of interelated elements. I've tried to build it up sequentially but sometimes you'll need to read further along to understand why something is the way it is.
+
+### The basics & task integration
```
chrisrpatterson: I don't think this is going to be correct in many cases. For the most part you will likely still want to run the task to normalize the cache to make sure you have the right versions as there are a lot of cases where you might choose a key like branchname and that would give you mostly the right files and then you only have to download the delta from the package source.

I also don't consider this to be a generic feature of tasks, I think having the ability to simply cache a set of files by a key and restore that is the baseline functionality and could work equally well for scenarios like build outputs as well as package cache pre-population

design/pipeline-caching.md
```diff
@@ -0,0 +1,366 @@
...
+
+From this screenshot you can see five inputs to the caching infrastructure:
+
+1. **Use cache flag**; controls whether the cache is used for this task in the case of a cache hit.
+2. **Populate cache flag**; controls whether this task is capable of populating the cache.
+3. **Key expression**; an expression that ultimately returns a string-based key which forms the key to identify the cache contents. May be combined with other scoping information before use.
+4. **Capture expression**; an expression that returns a set of things to capture. The most common case would be files, but it may also be necessary to capture things like environment variables etc which might be the output of a task.
+5. **Strategy**; an option that controls how cached content will be stored. For large artifacts our dedup protocol might be most appropriate, in other cases using a zip file might be more efficient.
+
+#### Cache use vs. population
```
chrisrpatterson: This makes sense though it should be controlled by simply using conditionals to not run the cache restore step in whatever cases you require.

design/pipeline-caching.md
```diff
@@ -0,0 +1,366 @@
...
+  version: 8
+- script:
+    | npm ci
+    | npm run build
+```
+
+Caching will work in a similar way. The following example sets up caching for all subsequent steps. Cache hit checking is evaluated on a per step basis and the step itself, and its order in execution forms part of the cache key (implicitly).
+
+```yaml
+- use: cache
```
chrisrpatterson: I see cache as it's own directive and not part of the use construct.

design/pipeline-caching.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Proposal
+
+This proposal contains a number of interelated elements. I've tried to build it up sequentially but sometimes you'll need to read further along to understand why something is the way it is.
+
+### The basics & task integration
+
+When work is done on an agent as part of a pipeline, it is represented as a _job_ when a set of _steps_ which are implemented by tasks. For many pipelines the YAML syntax obscures the underlying "tasky" nature of Azure Pipelines, but they are there nonetheless. As a result I'm starting with how pipeline caching will integrate with the underlying task infrastructure.
+
+Detecting a cache hit or miss will be performend by the agent worker. As the worker prepares to execute each task it will check to see whether any given task has caching enabled and if so it will determine the key and check for a cache hit. If a hit is detected it will set about restoring the cached content into the pipeline execution environment. Exactly how that content is restored will be discussed a bit later.
```
chrisrpatterson: No the agent or worker should not have any knowledge of the cache, it should be a step that runs.  This type of caching is best effort and getting a miss on the cache should not cause the pipeline to fail

PR #58, opened 2018-11-19
Trigger CI only on 'master' so that CI and PR builds don't both run for PR changes

DavidStaheli:
Let's say you use one of our YAML templates to add azure-pipelines.yml to your repo.  Later on, someone creates a new PR.  For every change committed to that PR, you'll get both a CI and a PR build.  It's because if no CI trigger is specified in the YAML, CI fires for every branch - PR or not.

This PR adds to each YAML template an explicit CI trigger for 'master' .  This results in users getting PR builds, and upon merging - a CI build.  But not both at the same time for a PR commit.

### DIFF
--- a/templates/.net-desktop.yml
+++ b/templates/.net-desktop.yml
@@ -3,6 +3,9 @@
 # Add steps that publish symbols, save build artifacts, and more:
 # https://docs.microsoft.com/azure/devops/pipelines/apps/windows/dot-net
 
+trigger:
+- master
+
 pool:
   vmImage: 'VS2017-Win2016'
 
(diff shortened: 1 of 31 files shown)


vtbassmatt:
Should we go ahead and configure the `pr` directive as well? Today it looks like we create new pipelines with an override by default - but I assume that behavior will change in the future. Othewise, LGTM.

DavidStaheli:
> Should we go ahead and configure the `pr` directive as well

Let's look into it.  I think one reason we create new pipelines with an overriden PR trigger is to enable the checkbox: "Build pull requests from forks of this repository"

image

PR #60, opened 2018-11-20
Multiple checkout spec

vtbassmatt:


design/multicheckout.md
```diff
@@ -0,0 +1,118 @@
...
+## Schema
+
+```yaml
+- checkout: none | self | resourceName
+  clean: boolean      # whether to fetch clean each time
+  fetchDepth: number  # the depth of commits to ask Git to fetch
+  lfs: boolean        # whether to download Git-LFS files
+  submodules: true | recursive  # set to 'true' for a single level of submodules or 'recursive' to get submodules of submodules
+  persistCredentials: boolean   # set to 'true' to leave the OAuth token in the Git config after the initial fetch
+  root: string        # directory to checkout the repo
```
TingluoHuang: what about `path` instead of `root`? that was the name we picked half year ago. :D

design/multicheckout.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+## New features needed
+
+- Resources directory: $(Pipelines.ResourcesDirectory)
+- Default source directory: $(Pipelines.DefaultSourcesDirectory)
+- `self` checkout influences $(Pipelines.SourcesDirectory)
```
TingluoHuang: why `self` influences `$(Pipelines.SourcesDirectory)`? it would make more sense to me it influences `$(Pipelines.DefaultSourcesDirectory)`

TingluoHuang: what's the different between these 2 folders.
here is the the agent work folder layout looks like:
```
D:\_work\1\s
                 \s\self
                 \s\toolRepo
                 \s\scriptRepo
```
which folder is $(Pipelines.ResourcesDirectory) and which folder is $(Pipelines.DefaultSourcesDirectory)?

design/multicheckout.md
```diff
@@ -0,0 +1,118 @@
...
+    name: CodeRepo
+    # TODO: how does a push to CodeRepo trigger this pipeline?
+
+steps:
+- checkout: none  # Pipelines.SourcesDirectory will be unset
+- checkout: code
+  root: $(Pipelines.DefaultSourcesDirectory)
+```
+
+### Code split up across multiple repos
```
TingluoHuang: why not using submodules? :D

design/multicheckout.md
vtbassmatt: That's fine by me.

design/multicheckout.md
vtbassmatt: ResourcesDirectory would be a new concept. Maybe `d:\_work\1\r`. By default, self goes in `d:\_work\1\s` and a repo called `foo` goes in `d:\_work\1\r\foo`.

DefaultSourcesDirectory is static - on Hosted, always `d:\_work\1\s`. SourcesDirectory is the actual place where your `self` repo was placed, so in your example, `d:\_work\1\s\self`. In a single-checkout case with no specified `path`, the two folders are identical.

If this design works, I'll clarify in the spec. If you're asking for changes, what do you propose?

design/multicheckout.md
vtbassmatt: Submodules have their place, but some customers have other ways of stitching together multiple repos. At Microsoft, for example, the Skype team uses dozens of repos and doesn't use submodules. I'm not convinced whether submodules are truly [evil](http://slopjong.de/2013/06/04/git-why-submodules-are-evil/), but multi-repos are not a solved/settled problem.

design/multicheckout.md
chrisrpatterson: If repo self goes to d:\_work\1\s then why wouldn't foo go to d:\_work\1\foo?  I am not sure I would put repos in the resources folder as that will simply make things difficult to use for the primary scenario of needing multiple repos for a build.

design/multicheckout.md
```diff
@@ -0,0 +1,118 @@
...
+resources:
+  repositories:
+  - repository: tools
+    type: git
+    name: ToolsRepo
+
+steps:
+- checkout: self
+- checkout: tools
+  root: buildTools  # relative paths are assumed to be rooted at $(Pipelines.ResourcesDirectory)
```
chrisrpatterson: Should this be ```path```?  root seems odd for some reason.

design/multicheckout.md
chrisrpatterson: Requiring someone to adopt submodules just to support the scenario of having YAML in a different repo is overkill and it puts a huge burden on requiring developers to understand a complex and not well supported feature of Git.

design/multicheckout.md
vtbassmatt: I'm not married to either of these.

I chose the resources directory for non-`self` repo to match what other resources will do. What do we do if the user starts giving repo resources names like "a"? I figured we own the top dirs (_work\\1\\r, _work\\1\\s, etc.) and then the user has an empty namespace inside to work with.

`path` vs `root` - no strong opinion here, but `path` does kinda read better. I'll update.

design/multicheckout.md
vtbassmatt: I chose the resources directory for non-`self` repo to match what other resources will do. What do we do if the user starts giving repo resources names like "a"? I figured we own the top dirs (_work\\1\\r, _work\\1\\s, etc.) and then the user has an empty namespace inside to work with.

vtbassmatt: I updated the spec. Now, I propose we put everything under the current `\s` sources directory in its own directory by default.

PR #67, opened 2018-12-03
upload/download keywords for Pipeline Artifacts

alexmullans:


design/pipeline-artifacts.md
```diff
@@ -1,30 +1,30 @@
...
 
-Artifacts are distinct from other `resources` types, including `containers`, `repositories`, and `feeds`.
+Artifacts are distinct from other `resources` types, including `containers`, `repositories`, `packages`, and `feeds`.
 
-## Publishing artifacts: `artifact`
+## Uploading artifacts: `upload`
 
-`artifact` is a shortcut for the [Publish Pipeline Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-pipeline-artifact.md) task. It will publish files from the current job to be used in subsequent jobs or in other stages or pipelines.
+`artifact` is a shortcut for the [Upload Pipeline Artifact](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-pipeline-artifact.md) task (formerly called "Publish Pipeline Artifacts"). It will upload files from the current job to be used in subsequent jobs or in other stages or pipelines.
```
vtbassmatt: Should be `upload`

design/pipeline-artifacts.md
```diff
@@ -128,22 +133,22 @@ You can add files to both the default artifact and to named artifacts until the
...
+  - upload: bin/WebApp/*
     name: WebApp
-  - artifact: bin/MobileApp/*
+  - upload: bin/MobileApp/*
     name: MobileApp
 - job: Deploy
   steps:
-  - downloadArtifact: WebApp
-  - downloadArtifact: MobileApp
+  - download: WebApp
```
ashokirla: In this example WebApp and MobileApp are pipeline artifacts and not resources. So the schema has to be 
-download: 
  name: WebApp
-download: 
  name: MobileApp

design/pipeline-resources.md
```diff
@@ -27,73 +27,63 @@ resources:
...
 If you have a pipeline that produces artifacts, you can consume the artifacts by defining a `pipelines` resource. A pipeline can be another Azure DevOps pipeline or any external pipelines like Jenkins etc.
 
-
 ### Schema
 
 ```yaml
 resources:        # types: pipelines | repositories | containers | packages
   pipelines:
   - pipeline: string  # identifier for the pipeline resource
-    type: enum  # type of the pipeline source like azurePipelines, Jenkins etc.
```
ashokirla: In the schema definition, we still have to show the input `type`. For azure pipelines, this input is not required. But for other pipeline services like Jenkins this is required.

vtbassmatt: Per an offline conversation: we'll keep around the notion of an extensible category for CI systems. It won't be "pipeline" -- that's reserved for Azure Pipelines.

vtbassmatt: @ashokirla do you want to take a run at keywords for this, or do you want me to? I like the idea of "pipeline/upload/download" -> Azure Pipelines, with something else representing other CI/orchestration systems.

design/pipeline-artifacts.md
```diff
@@ -1,30 +1,30 @@
...
 
-Artifacts are distinct from other `resources` types, including `containers`, `repositories`, and `feeds`.
+Artifacts are distinct from other `resources` types, including `containers`, `repositories`, `packages`, and `feeds`.
 
-## Publishing artifacts: `artifact`
+## Uploading artifacts: `upload`
 
-`artifact` is a shortcut for the [Publish Pipeline Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-pipeline-artifact.md) task. It will publish files from the current job to be used in subsequent jobs or in other stages or pipelines.
+`upload` is a shortcut for the [Upload Pipeline Artifact](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-pipeline-artifact.md) task (formerly called "Publish Pipeline Artifacts"). It will upload files from the current job to be used in subsequent jobs or in other stages or pipelines.
```
ericsciple: what happens on retry? does it potentially dirty the same artifact? will it overwrite previously uploaded files?

design/pipeline-artifacts.md
```diff
@@ -33,20 +33,19 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
 
 ---
 
-## Downloading artifacts: `downloadArtifact`
+## Downloading artifacts: `download`
 
-`downloadArtifact` is a shortcut for several tasks:
+`download` is a shortcut for two tasks:
```
ericsciple: What is the experience when a pipeline does not have any artifacts published to it? Do both steps get added anyway? If so, should we combine the logic into a single step instead?

The nice thing about always showing the task, is that it advertises the feature. However it seems silly for a single-job pipeline.

Is it really only two tasks? Aren't artifacts extensible? What about when downloading artifacts from a pipeline with extensible artifact types published to it?

design/pipeline-artifacts.md
```diff
@@ -33,20 +33,19 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
 
 - The [Download Pipeline Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact) task
 - The [Download Fileshare Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-fileshare-artifacts) task
-- Download tasks for other pipelines systems e.g. Jenkins
 
-It will download artifacts published from a previous job or stage or from another pipeline. Artifacts are downloaded either to `$PIPELINES_RESOURCESDIR` or to the directory specified in `root`.
+It will download artifacts uploaded from a previous job or stage or from another pipeline. Artifacts are downloaded either to `$PIPELINES_RESOURCESDIR` or to the directory specified in `root`.
```
ericsciple: What is `$PIPELINES_RESOURCESDIR`? Is this a new folder under the build directory? What is the folder name? `r`?

Have you considered the singular form? `PIPELINE_RESOURCESDIR`?

All other directory varibles spell out the full word "DIRECTORY" (except for `agent.workFolder`). Shouldn't we be consistent here and name the variable `pipeline.resourcesDirectory`?

design/pipeline-artifacts.md
```diff
@@ -1,30 +1,30 @@
...
 
-In this document, `artifact` refers specifically to publishing Pipeline Artifacts from the current pipeline. `downloadArtifact` refers to downloading Pipeline Artifact artifacts from the current pipeline, from other Azure Pipelines, and from other pipeline resources (e.g. Jenkins builds). 
+In this document, `artifact` refers specifically to uploading Pipeline Artifacts from the current pipeline. `download` refers to downloading Pipeline Artifact artifacts from the current pipeline and from other Azure Pipelines. Other pipeline systems (e.g. Jenkins) will be handled as [pipeline resources](pipeline-resources.md).
 
-Artifacts are distinct from other `resources` types, including `containers`, `repositories`, and `feeds`.
+Artifacts are distinct from other `resources` types, including `containers`, `repositories`, `packages`, and `feeds`.
 
-## Publishing artifacts: `artifact`
+## Uploading artifacts: `upload`
```
ericsciple: to be clear: we are not adding an alias for the pulish-file-share-artifacts task, correct?

design/pipeline-artifacts.md
```diff
@@ -95,31 +100,31 @@ You can control the location where artifacts are downloaded using the `downloadR
 - job: Build
   steps:
   - script: dotnet publish --configuration $(buildConfiguration)
-  - artifact: bin/*
+  - upload: bin/*
 - job: Deploy
   steps:
-  - downloadArtifact:
+  - download:
```
ericsciple: What is the default root? When referencing a pipeline resource, is it \<PIPELINE_ALIAS\>/\<ARTIFACT_NAME\>?

What about for the current pipeline? Is it just the \<ARTIFACT_NAME\>? And we don't need to prefix with `self` or anything because the likelyhood of collision is low?

design/pipeline-artifacts.md
ericsciple: What if you specify a relative path? Would it be rooted against the System.DefaultWorkingDirectory (i.e. Build.SourcesDirectory in build)? That is how filePath inputs work today. Or is this a regular string input?

design/pipeline-artifacts.md
```diff
@@ -33,20 +33,19 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
-- Download tasks for other pipelines systems e.g. Jenkins
 
-It will download artifacts published from a previous job or stage or from another pipeline. Artifacts are downloaded either to `$PIPELINES_RESOURCESDIR` or to the directory specified in `root`.
+It will download artifacts uploaded from a previous job or stage or from another pipeline. Artifacts are downloaded either to `$PIPELINES_RESOURCESDIR` or to the directory specified in `root`.
 
 ### Schema
 
 ```yaml
-- downloadArtifact: string # identifier for the resource from which to download artifacts, optional; defaults to 'self`
+- download: string # identifier for the pipeline resource from which to download artifacts, optional; defaults to `self`, which represents the current pipeline
```
ericsciple: Is it weird to use `self` here? In a yaml build, `self` is the repo where the entry yaml file is defined. Would it be better to say `""` means the current pipeline rather than overload the concept `self`?

design/pipeline-resources.md
```diff
@@ -27,73 +27,63 @@ resources:
...
 
 ```yaml
 resources:        # types: pipelines | repositories | containers | packages
   pipelines:
   - pipeline: string  # identifier for the pipeline resource
-    type: enum  # type of the pipeline source like azurePipelines, Jenkins etc. 
-    connection: string  # service connection to connect to the source
-    project: string # project for the source; optional for current project  
+    connection: string  # service connection for pipelines in other Azure DevOps organizations or Servers
+    project: string # project for the source; optional for current project
```
ericsciple: We need to change `version` to `versionSpec`. When the plan is created, we need to resolve the pipeline resource to a specific version. "Latest" is not a "version". "Latest" is a version spec. It resolves to a specific build ID.

We need to stamp the resource with the resolved version, and store that in the generated plan environment.

All stages and jobs must resolve to the same version.

ericsciple: And retried nodes should resolve to the same version (part of the same plan).

I would like to see retry/rerun behaviors called out explicitly. What happens during requeue?

design/pipeline-resources.md
```diff
@@ -27,73 +27,63 @@ resources:
...
 ```
 
+Refer to [download artifacts](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md#downloading-artifacts-download) for more details.
 
-Refer to [download artifacts](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md#downloading-artifacts-downloadartifact) for more details.
-
-Artifacts from the `pipeline` resource are downloaded to `$PIPELINES_RESOURCESDIR/<pipeline-identifier>/` folder.
+Artifacts from the `pipeline` resource are downloaded to `$PIPELINES_RESOURCESDIR/<pipeline-identifier>/<artifact-identifier>` folder.
 
 ## Resources: `repositories`
```
ericsciple: Repository does not support a `refs` property, it should be `ref`

design/pipeline-resources.md
ericsciple: Have you considered moving branch and tags under versionSpec? Something like
```yaml
- pipeline: alias
  project: ""
  source: ""
  versionSpec:
    quality: succeeded
    ref: master
    tags: ""
```

If not, what are the values for `versionSpec` other than `Latest`?

Consider `ref` instead of `branch` (match repository)

Is `tags` a sequence? Or is it a string? If it is a sequence, are they AND'd?

How are you envisioning adding trigger specifications? With a `trigger` property, and then everything relevant nested?

design/pipeline-resources.md
```diff
@@ -27,73 +27,63 @@ resources:
...
-Artifacts from the `pipeline` resource are automatically downloaded and made available for all the jobs. However, in any of the jobs, you can choose to override and download only specific artifacts using `downloadArtifact` shortcut. Once you use `downloadArtifact` to override and download a specific artifact, automatic artifact download behavior is removed and you need to specify all the artifacts you intend to download in the job.
+### `download` for pipelines
 
+All artifacts from the current pipeline and from all `pipeline` resources are automatically downloaded and made available at the beginning of each job. You can override this behavior: see [Pipeline Artifacts](pipeline-artifacts.md#default-and-named-artifacts) for more details.
 
 ```yaml
 - job: deploy_windows_x86_agent
   steps:
-  - downloadArtifact: SmartHotel   # pipeline resource identifier.
+  - download: SmartHotel   # pipeline resource identifier.
```
ericsciple: Does the metadata for the referenced resource go down in the job message, to the agent? We can expose the metadata as variables or, even better so we dont bloat the variables, as expression context.

What is the longer term plan here?

design/pipeline-artifacts.md
ashokirla: There is a possibility that the internal artifact name "WebApp" can collide with pipeline resource name. 
And one input type is supposed to take one type of object. In this case `-download` will end up having two possible input types (pipeline resource or internal artifact name). This will be confusing. Alex and I discussed about this when he was in IDC.

design/pipeline-resources.md
ashokirla: Yeah, i am working on the proposal for the new set of keywords for external CI systems.

design/pipeline-artifacts.md
vtbassmatt: Is the risk greater here than it was with containers? Here are two legal YAML files which do something quite different:

```yaml
# explicit resource
resources:
  containers:
  - container: python
    image: python:2.7-windowsservercore-ltsc2016

container: python  # a resource pointing to the Python 2.7 container on Windows
steps:
- script: echo foo
```

```yaml
# no explicit resource
container: python  # no resource found; this is the default tag for Python, probably a 3.7.x on Linux
steps:
- script: echo foo
```

But in both cases, these files do (likely) exactly what the user expects.

design/pipeline-artifacts.md
vtbassmatt: Or perhaps make it explicit but not use the keyword `self` -- how bad would it be to squat on `current`?

ericsciple: I wonder if we need to make resource aliases distinct across categories? If so, then this should not be `self`.

Also note, the name will matter when we expose resources into the expression context. Currently only resources are exposed and only for the extension injection stuff.

design/pipeline-resources.md
ashokirla: Exposing artifact meta-data as variables is one of the value proposition of Resources and it enables bunch of important scenarios. This need not be considered as bloating of variables list. Today, in RM, we expose bunch of build artifact meta-data as variables like source definition, build number, commit sha and few more. Customers use this info as part of their deployment workflow, conditional tasks and triggers.

design/pipeline-resources.md
ashokirla: i have updated the PR with the suggested changes to the spec.

alexmullans: Great, resolving this conversation since you've updated the PR.

design/pipeline-artifacts.md
```diff
@@ -1,30 +1,30 @@
...
-  prependPath: string # a directory path that will be prepended to all published files
+  prependPath: string # a directory path that will be prepended to all uploaded files
   seal: boolean # if true, finalizes the artifact so no more files can be added after this step
 ```
 
 ### Example
 
 ```yaml
-- artifact:
+- upload:
```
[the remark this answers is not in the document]
alexmullans: We considered that option but went with what's currently in the spec as it allows for a one-liner in the simple case: `- upload: <my file selector>`. (This would upload files to the `default` artifact.) I'm open to discussions on consistency though. @vtbassmatt your thoughts?

design/pipeline-artifacts.md
alexmullans: Yes, unless the user chooses to seal, they can update the artifact multiple times. This includes overwriting files uploaded in the first try or adding new files that weren't present in the first try.

design/pipeline-artifacts.md
alexmullans: Since we're limiting this syntax to only pipeline artifacts, I should probably change this to say that `download` is strictly a shortcut for the Pipeline Artifacts task. Users needing to publish artifacts to/download artifacts from file shares would directly reference the appropriate tasks, rather than using the `upload`/`download` shortcuts. Thoughts?

design/pipeline-artifacts.md
alexmullans: I'd pulled those names from Matt's [branch on variables](https://github.com/Microsoft/azure-pipelines-yaml/blob/proposals/variables/design/variables-in-pipelines.md). I don't want to define a new directory here - my intent is to use the existing directory. I'm also happy to align on variable names. So $(Pipeline.ResourcesDirectory) or $PIPELINE_RESOURCESDIRECTORY then?

design/pipeline-artifacts.md
alexmullans: Correct.

design/pipeline-artifacts.md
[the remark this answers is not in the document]
ericsciple: That's fine if you want to optimize for the selector. But I do think the property that denotes the artifact name should be `artifact` or `artifactName`. The problem is `name` is a core property on almost all of the steps. It means the "task reference name" and is used when programmatically accessing the output variables set by the task.

design/pipeline-artifacts.md
alexmullans: Great question. Clarified on Line 43.

design/pipeline-artifacts.md
alexmullans: I like leaving it blank - changed.

design/pipeline-resources.md
alexmullans: Leaving this feedback for @ashokirla

alexmullans: Leaving this for @ashokirla as well.

alexmullans:
@ericsciple @vtbassmatt @ashokirla sorry for the delay, I've been out recovering from surgery. I've made some fixes and responded to  the comments above - can you take another pass and resolve conversations if you feel I've addressed your feedback?

design/pipeline-artifacts.md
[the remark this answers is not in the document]
alexmullans: Ah, makes total sense. Changed to `artifact`.

design/pipeline-artifacts.md
ericsciple: Has consistency been considered? In other words, should we allow `checkout: self` to be consistent?

Personally, I haven't heard users complain about `checkout: self`. And generally I think we confuse users when we have multiple ways of doing the same thing.

design/pipeline-artifacts.md
ericsciple: This is not my understanding. It might be good to touch base for a few minutes later today. I've talked to multiple people and heard multiple answers.

design/pipeline-resources.md
ashokirla: The version here means the build number. By default, it will be resolved to latest successful pipeline run (across all stages).
If user wants to pick the latest from a specific branch or tag, user can specify the branch/tag inputs in the definition. Then we will fetch the latest successful pipeline from given branch/tag.

However, if user choses to consume a specific version, then he/she can give the pipeline run number as a 'version' input. In that case that version is picked. 

We will not consider any failed/cancelled pipeline runs while resolving the version. We will start with this simpler experience. Later if we see any valid scenarios for consuming artifacts from a failed pipeline, we can extend by introducing 'quality' and 'stage' as additional inputs as you have suggested.

And yes, we stamp the resource with the resolved version for any pipeline run instance. And the retry/redeploy scenarios also will use the same version. This is the same behavior we have in RM while resolving build artifacts. 

I am happy to get on a call if you have any concerns/feedback.

ashokirla: We initially started with that design by keeping branch, tags, version under one node. I discussed this with Chris and Matt. Since these are anyway pipeline resource specific properties, the additional depth is not required. We can achieve all the scenarios by having version, branch and tag inputs at the same level. 

Regarding 'branch' vs 'ref', a pipeline can be associated with any kind of version control system. So we think keeping it generic as 'branch' would be more perceivable to user :).

Tag is a sequence i.e. list of strings. If multiple tags are given, they are AND'd. 

Triggers will be a separate property in pipeline resource and all the trigger specific properties will be nested under that node. I will share a separate spec for Triggers.

We will consider latest successfully completed pipeline run. This means all the stages of the pipeline should be succeeded or partially succeeded.  Like I explained above, we will start with this and then based on the scenarios we can extend.

design/pipeline-artifacts.md
ashokirla: Yeah, I realize we follow the same pattern for containers.

With templates, there is very good possibility that build stage where artifacts are published can be defined in some other file. And user wouldn't know what is the artifact name given while publishing unless he/she verifies all the files carefully. 

Moreover, this also means there are two different ways of doing the same thing:
-download:
  name: WebTier
vs
-download: WebTier

And the same input field taking two different types of objects (resource and artifact) can also be confusing.

vtbassmatt: OK, I think I conflated two things. After reading more carefully, I agree with @ashokirla here.

design/pipeline-artifacts.md
vtbassmatt: +1 to @ericsciple's point, I don't like blank implying a behavior.

design/pipeline-artifacts.md
vtbassmatt: Sorry Eric, I caused the confusion here. We need to get aligned on variables, and whatever we align behind is what artifacts will use. Let me continue that conversation in the other PR.

design/pipeline-artifacts.md
vtbassmatt: Context for my conversation with Alex: can one instance of `download` always resolve to a single task? I thought that's what we said in the PM spec, but apparently that's not what's being implemented right now.

design/pipeline-artifacts.md
vtbassmatt: Not sure this clarified it for me. Eric pointed out that although you can't have two artifacts with the same name in "this" pipeline, you certainly can as soon as you start pulling in other pipelines. We'd be smarter/better off to simply namespace all the artifacts (using folders named for the pipeline) from the get-go. This will be a breaking change (or at least needlessly complicated and confusing) if we introduce it later.

design/pipeline-resources.md
vtbassmatt: I think everyone's in violent agreement that we need to do this :) The question is the mechanism. It would be nice to hear if there are customer scenarios where auto-injected variables are *required*, or if we can put this into expression context and let customers *opt in* to variable injection.

design/pipeline-artifacts.md
alexmullans: Great. I'll close this conversation for now.

design/pipeline-artifacts.md
alexmullans: Yes, it should be a single task, as agreed.

design/pipeline-artifacts.md
alexmullans: Chatted IRL - we'll keep this design for now, and assume that likelihood of collision is low.

design/pipeline-resources.md
alexmullans: Closing this discussion here for now. Please feel free to continue it in a dedicated Pipeline Resources PR.

alexmullans:
Merging after final review with @vtbassmatt (covering @ericsciple's feedback).

PR #69, opened 2018-12-05
"downloadRoot" is "root"

chcosta:


vtbassmatt:
Fixed in another PR - thanks Chris!

PR #82, opened 2018-12-14
Triggers in resources.

ashokirla:
Introduced triggers for all resources in YAML.


vtbassmatt:
(I shared these in chat, but putting them out here publicly as well.)

I'd like to start with customer scenarios. I don't have enough context about the customer problem and existing solutions, etc. to appropriately judge a design. These functional requirements are helpful but not enough.

Also, there's been some confusion about the keys under `resources`. My belief is that we need to make *that* extensible, so that instead of:
```yaml
resources:
  builds:
  - build: foo
    type: jenkins
```
instead we can directly offer:
```yaml
resources:
  jenkins:
  - definition: foo  # or whatever key makes sense here
```

vtbassmatt:
(These are new, not the ones I shared via chat.)

It would also help to begin with a short, functional discussion of the general problem "triggers" solve. (Likewise "tracking", once we get to that concept.)

How should we think about the existing, pipeline-wide `self` repository triggers? They offer a lot of convenience for simple cases, but aren't flexible enough for multi-stage pipelines.

My sense is that we need to say more about the variables and context available to task authors and scripters, too. That's not strictly a concern of triggers, but it does get faulted in. Today, depending on the exact trigger, we inject things like:
- Build.TriggeredBy.BuildId
- Build.Reason
- System.PullRequest.PullRequestId

and I imagine RM has some of the same.

vtbassmatt:
@ericsciple @patrickcarnahan @bryanmacfarlane please take a look when you can. @ashokirla is going to factor out triggers into its own doc for clarity.

ashokirla:
Thanks for the comments @vtbassmatt. 

Regarding the categories under `resources`, I will get back to you with the customer scenarios and how we can achieve them with the design we currently closed on and also evaluate the alternatives. 
However, I would like to have that discussion separate so that we can close on trigger spec for pipelines, repositories and containers which is our priority now. 

I will move the triggers spec into a separate doc and also share the scenarios we are going enable with triggers as suggested. 

And yes, we push resource variables including trigger specific variables which can be consumed by pipeline stages, jobs and tasks. In RM today, we provide the details about the artifact source that triggered (`Release.TriggeringArtifact.Alias`) the release and bunch of source specific info including the version, commit sha etc. I will mention about this in the doc.

ashokirla:
Created another PR for Triggers: https://github.com/Microsoft/azure-pipelines-yaml/pull/91

PR #83, opened 2018-12-14
artifacts suggestions

ericsciple:


design/pipeline-artifacts.md
```diff
@@ -15,9 +15,9 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
 ### Schema
 
 ```yaml
-- upload: string | [ string ]  # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to upload
+- upload: string # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to upload
```
ericsciple: I'm proposing we switch this to `string` only, not `string | [ string ]`.

Reasoning: I'm working on changing the parsing infrastructure of the agent to be schema driven - for extensibility. Ultimately this property maps to a task input, which is a `string` not an array of strings. And we don't have any mechanism in the parser currently to convert an array of strings into a single string. That's not to say we can't add it in the future, I just want to eliminate adding a new concept to solve it right now. And a multi-line string is allowed already.

design/pipeline-artifacts.md
```diff
@@ -15,9 +15,9 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
   artifact: string # identifier for this artifact (no spaces allowed), defaults to 'default'
-  prependPath: string # a directory path that will be prepended to all uploaded files
+  prependDirectory: string # a directory path that will be prepended to all uploaded files
```
ericsciple: Minor suggestion of calling this `prependDirectory` to be clear. I don't feel too strongly about this, but I do think it aligns better with our in-the-box variable names.

When I read the spec, I saw the examples all ended with a trailing `/` to make it clear that it's a directory. And then I had to read the description to understand what the expectation is, if it doesn't end with a trailing `/`.

So my gut feel is, why not name it more intuitively?

Again, just a suggestion. I dont feel too strongly about this.

design/pipeline-artifacts.md
```diff
@@ -47,9 +47,16 @@ Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the dire
 
 ```yaml
 - download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline
-  artifact: string # identifier for the artifact to download; if left blank, downloads all artifacts associated with the resource provided
-  patterns: string | [ string ] # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to download; if blank, the entire artifact is downloaded
+  artifact: string # identifier for the artifact to download; required
```
ericsciple: Without making the `artifact` property required, it introduces confusing rules about how `downloadPath` should work, when multiple artifacts are resolved.

Otherwise, how should it work? And here are the examples that would need to be answered:

For the current pipeline, with multiple artifacts:

```yaml
jobs:
- job: a
  steps:
  - upload: "*.dll"
    artifact: bin
  - upload: "*.pdb"
    artifact: pdb
- job: b
  dependsOn: a
  steps:
  - download: ""
# I would expect this to translate to two tasks:
# - task: …
#    downloadPath: bin
# - task: …
#    downloadPath: pdb
```

For a referenced pipeline:

```yaml
resources:
  pipelines:
  - pipeline: other # let's say it has two artifacts, "bin" and "pdb"
steps:
- download: other
# I would expect this to translate to two tasks:
# - task: …
#    downloadPath: other/bin
# - task: …
#    downloadPath: other/pdb
```

For the current pipeline with multiple artifacts, and a specified `root`:

```yaml
jobs:
- job: a
  steps:
  - upload: "*.dll"
    artifact: bin
  - upload: "*.pdb"
    artifact: pdb
- job: b
  dependsOn: a
  steps:
  - download: ""
    root: asdf
# I would expect this to translate to two tasks:
# - task: …
#    downloadPath: asdf/bin              # IS THIS RIGHT?
# - task: …
#    downloadPath: asdf/pdb
```

For a referenced pipeline, and a specified `root`:

```yaml
resources:
  pipelines:
  - pipeline: other # let's say it has two artifacts, "bin" and "pdb"
steps:
- download: other
   root: asdf
# I would expect this to translate to two tasks:
# - task: …
#    downloadPath: asdf/bin            # IS THIS RIGHT?
# - task: …
#    downloadPath: asdf/pdb
```

For the current pipeline, a specific artifact, and a specified `root`:

```yaml
jobs:
- job: a
  steps:
  - upload: "*.dll"
    artifact: bin
  - upload: "*.pdb"
    artifact: pdb
- job: b
  dependsOn: a
  steps:
  - download: ""
    artifact: bin
    root: asdf
# I would expect this to translate to one task:
# - task: …
#    downloadPath: asdf              # IS THIS RIGHT?
```

design/pipeline-artifacts.md
```diff
@@ -47,9 +47,16 @@ Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the dire
...
+  patterns: string # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to download; if blank, the entire artifact is downloaded
   root: string # the directory in which to download files, defaults to $PIPELINE_RESOURCESDIRECTORY; if a relative path is provided, it will be rooted from $SYSTEM_DEFAULTWORKINGDIRECTORY
+  displayName: string # friendly name displayed in the UI
+  name: string # identifier for this step (A-Z, a-z, 0-9, and underscore)
+  condition: string
+  conditionOnError: boolean # 'true' if future steps should run even if this step fails; default to 'false'
+  enabled: true # whether or not to run this ste; defaults to 'true'
+  timeoutInMinutes: number
```
ericsciple: These are core step properties. Can they all be set as well?

Note, this is also relevant to whether we change `artifact` to be required. If not, then the `timeout` property doesn't make sense when the single `download` step expands to multiple steps.

design/pipeline-artifacts.md
```diff
@@ -47,9 +47,16 @@ Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the dire
 
 ```yaml
 - download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline
```
ericsciple: I think the pipeline alias `current` makes sense. This is a good time to say, resource aliases must be unique. And we should add the validation along with this feature. This gives us the flexibility in the future to expose resources as one big dictionary (not necessarily in yaml, but in expression context).

This is at least a good time to ask the question. I'm a little bit torn whether we need to do this, but leaning in the direction of buying us future flexibility.

Thoughts?

design/pipeline-artifacts.md
vtbassmatt: Agreed, let's do it.

design/pipeline-artifacts.md
vtbassmatt: This one I can take or leave. I've used `path` elsewhere even when it's expected to be a directory. Partly, it's easier to spell 😆 But I don't mind "directory" either, as it's more precise.

vtbassmatt: The main argument in favor of `path` is that it's abstract from the OS. For instance, we can say that only forward slashes are accepted as separators (and do the translation on the user's behalf, under the covers, if they're on Windows). Whereas with `directory`, a Windows user is going to expect to be allowed to use backslashes. Not a super strong argument.

design/pipeline-artifacts.md
vtbassmatt: (I don't have a strong opinion here, other than we should be predictable and flexible.)

I was nodding along to all your examples until the last one. I _think_ that should translate to
```yaml
- task: ...
  inputs:
    downloadPath: asdf/bin
```

@ashokirla, thoughts?

design/pipeline-artifacts.md
vtbassmatt: I think line 56 should be "continueOnError". I'm in favor of adding these core properties, but it's low priority.

design/pipeline-artifacts.md
vtbassmatt: After our conversations in December, I tend to agree. I think Patrick had some concerns (not sure what they were), did you and he get on the same page?

design/pipeline-artifacts.md
alexmullans: Yep, this was me not fully understanding YAML. As long as multiple strings can be provided, I'm 👍.

design/pipeline-artifacts.md
alexmullans: I'll let @vtbassmatt and @ericsciple make the final call, since you two have a better view of priorities (e.g. consistency vs. precision) across the whole YAML schema.

design/pipeline-artifacts.md
alexmullans: 👍

design/pipeline-artifacts.md
ericsciple: My point here, is actually that the artifact property should be `required`. Then you eliminate the confusion I described.

vtbassmatt:
@mitchdenny thanks for reading through.

**Task rename:** assuming we are going to major-version the task anyhow (from 0 -> 1), changing the name isn't really any extra effort for YAML users (though, I suppose it's one thing they could accidentally omit). I'm less sure about designer users. I'll find out.

**Collapsing sequential tasks:** that's not something we have today. I can't think of any other tasks that would want it, either, so I'm not sure we want to bake it into the platform. I thought about some creative solutions involving abusing the pre-job steps feature, but that breaks the semantics e.g. when you insert a download in the middle of the job. Two thoughts: 1) can you quantify the difference in some way, so we know what we're passing up if we don't do this? 2) could we ship without it and potentially add it later?

design/pipeline-artifacts.md
ashokirla: I think it is more of a convenience rather than confusion. 

If we look at the scenarios. 
Scenario 1: User would like to download all the artifacts from `pipeline` resource to a specific destination folder.
Scenario 2: User would like to download files with a given pattern 'SmarHotel**.tar.gz' across all the artifacts from the `pipeline` resource.

These are very valid scenarios. And by making the artifact input optional,
1. User doesn't have to define one section each for each of the artifact.
2. User doesn't have to be aware of the exact name of the artifact published. 
Here is one of the example RD today where user provides the file input from artifact without even knowing the name of the artifact.
image

Infact there is a feedback from the team to provide a default syntax that can work for both current pipeline and other pipeline resource together

i.e.
```yaml
download: current
root: blah
# This downloads artifacts from current pipeline to a specific destination.
```
```yaml
download: other
root: blah
# This downloads artifacts from 'other' pipeline to a specific destination
```
```yaml
download:
root: blah
#This download artifacts from both other and current pipeline(all the available azure pipelines).
```
We are now using the macro `download` only for azure pipelines. And we also decided to use the keyword 'current' for current pipeline. So we can reserve "" (empty) for referring both internal and external pipeline resources.

design/pipeline-artifacts.md
ashokirla: @vtbassmatt : Can you add this to the list of things you plan to support in v1 of the YAML. Somehow asking user to provide multi-line string for patterns doesn't feel like a good YAML experience :).

vtbassmatt: I'll add it for discussion. It feels like shades of the same "is there more than one syntax for things?" that we want to avoid except when it's unambiguously better for the vast majority of cases. Multiline input is well-trod ground in our tasks (both from the designer era and now in YAML).

design/pipeline-artifacts.md
mitchdenny: How common is the requirement to download a set of files across all artifacts? It seems like a corner case and how do you deal with collisions?

mitchdenny:
@vtbassmatt - I think we can defer it to later as an optimization. I don't have hard numbers yet but consider the following scenario. I have a build that produces 1GB of content, I want to upload that into multiple discrete artifacts - so I have multiple publish/upload steps. We need to hash 1GB for each one of those steps. Once you've hashed the upload might only take a few seconds because of the impact of deduplication but we still have to do the chunking each time to determine the delta.

design/pipeline-artifacts.md
vtbassmatt: I merged this PR, but I don't think `artifact` should be required. I'll put some examples in the PR I'm working on now.

PR #91, opened 2019-01-01
Pipeline triggers

ashokirla:
As suggested by Matt, in the PR: https://github.com/Microsoft/azure-pipelines-yaml/pull/82,  I have moved triggers spec to separate file and added scenarios. 

I will be using this PR for Triggers spec and remove the other PR.


design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+
+#### Schema
+```yaml
+resources:       
+  repositories:
+  - repository: string    
+    type: enum 
+    connection: string 
+    source: string  
+    trigger:  # Optional; Triggers are enabled by default
```
vtbassmatt: Making this enabled by default would be a breaking change. Also, for a non-`self` repository, I'm not sure we want to trigger by default. There are scenarios where the secondary repository shouldn't trigger (templates and external tools are two that come immediately to mind).

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+
+#### Schema
+```yaml
+resources:          
+  containers:
+  - container: string       
+    type: enum  
+    connection: string 
+    image: string # container image name, Tag/Digest is optional; defaults to latest image
+    trigger: # Optional; Triggers are enabled by default
```
vtbassmatt: Same as above, this would be a breaking change and there are solid scenarios where triggering isn't desired.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+    trigger: 
+      tags:
+        include: 
+        - production*
+      location: 
+      - east-US
+      - west-EU
+```
+
+## Schedule triggers
```
vtbassmatt: This is out of scope for resource triggers, but I understand why it's included for completeness. You might just say something to that effect.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
+# Triggers in pipelines
+
+Triggers are events on which you can start your pipeline run automatically. You can enable triggers on your pipeline by subscribing to both internal and external events.
+
```
vtbassmatt: This would be a good place to illustrate with some customer-focused scenarios (just a handful). Then, when we have questions or disagreements about a requirement below, we can use the customer voice as a starting point for resolving.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+You can enable triggers on the resources defined in your pipeline. Resources can be of types pipelines, repositories, containers and packages.
+Triggers are enabled by default on all the resources. However, you can choose to override/disable triggers for each resource.
+
+
+
+### Pipelines
+A new pipeline is triggered automatically whenever a new run of the `pipeline` resource is succesfully completed. See [pipeline resources](pipeline-resources.md#resources-pipelines) for more details.
+
+#### Scenarios
+- I would like to trigger my pipeline when a new ‘helm-package’ artifact is published by ‘Helm-CI’ pipeline that ran on `releases/*` branch.
```
vtbassmatt: This suggests you want to filter on produced artifact names, but I don't see anything like that below.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+    trigger:     # Optional; Triggers are enabled by default.
+      branches:  # branch conditions to filter the events, optional; Defaults to all branches.
+        include: [ string ]  # branches to consider the trigger events, optional; Defaults to all branches.
+        exclude: [ string ]  # branches to discard the trigger events, optional; Defaults to none.
+      stages: [ string ]  # trigger after completion of given stage, optional; Defaults to all stage completion.
+      tags: [ string ]  # tags on which the trigger events are considered, optional; Defaults to any tag or no tag.
+```
+
+#### Examples
+Triggers are enabled by default unless expliciltly opted out. You can disable the triggers on the `pipeline` resource.
```
vtbassmatt: I agree that for pipelines, automatic triggers make the most sense.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+```yaml
+resources:
+  pipelines:
+  - pipeline: SmartHotel
+    source: SmartHotel-CI 
+    trigger: none   
+```
+
+
+You can control which branches get the triggers with a simple syntax.
```
vtbassmatt: I'm (personally) of very mixed opinions on the simple syntax we have. The major downsides are 1) offering multiple right ways to say the same thing and 2) complicating the learning curve for "growing up" to more sophisticated uses. In a case like `self` repo branch include triggers, where that's far and away going to be the most common scenario, the upside of shorter syntax and the fact that it's reasonably unambiguous pays off that debt. But for pipeline triggers (and other types below), are we sure that's true? (In a nutshell: are you sure you want to have two syntaxes here?)

design/pipeline-triggers.md
ashokirla: If we look at CD scenarios, usually, the artifact name doesn't change from run to run. However what changes from run to run can be branch or tag etc.

design/pipeline-triggers.md
ashokirla: Sure, will add few scenarios.

design/pipeline-triggers.md
ashokirla: Sure, will do.

design/pipeline-triggers.md
ashokirla: Yeah, I am not comfortable with two syntaxes either. There are two reasons for providing simpler syntax:
1. Since we allowed simpler syntax for self repo, user tends to use the same syntax and expects it to work.
2. The simple syntax works for most of the scenarios.

design/pipeline-triggers.md
chrisrpatterson: We already have a container resource that can be used in a pipeline that does not trigger.  Having this automatically start triggering on existing pipelines one day is going to unexpected behavior for the users.  I would favor an explicit model here and we can certainly have our default getting started experience explicitly define the trigger.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+  containers:
+  - container: MyACR    
+    connection: RMPM
+    registry: contosodemo
+    image: Microsoft/alphaworz
+    trigger: 
+      tags:
+        include: 
+        - production*
+      location:
```
chrisrpatterson: As I understand it the location data is only part of the payload when we are using ACR and not when coming from DockerHub or Quay.  How are we thinking about addressing that issue?

design/pipeline-triggers.md
chrisrpatterson: I think it is reasonable to require the full syntax when dealing with triggers on resources.  From the perspective of what the customer is trying to do they are already much deeper in than the simple CI scenario where some of the brevity is driven by trying to compete with some of our simpler competitors.

design/pipeline-triggers.md
vtbassmatt: OK, that makes sense. Might want to update the scenario, then, to make it unambiguous what you're trying to support.

design/pipeline-triggers.md
vtbassmatt: Let's definitely discuss it, @ashokirla. I think a reasonable stance would be:
- Because of the scenario, we wish containers automatically came with triggers.
- Because of backcompat, we can't do that.
- In "v0" (unversioned) YAML, container triggers must be explicit. When we rev to v1 (we're going to eventually add a "version: int" field to the pipeline-level), we can change to the behavior we want.

I'll start a list of all the things we want to do with v1 syntax.

vtbassmatt: [Here is that list](https://github.com/Microsoft/azure-pipelines-yaml/pull/92).

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+ ```yaml
+resources:
+  pipelines:
+  - pipeline: SmartHotel
+    source: SmartHotel-CI 
+    trigger: 
+      branches:
+        include: 
+        - releases/*
+      tags:
```
ericsciple: are the tags and'd or or'd ? Can you clarify in the documentation?

design/pipeline-triggers.md
```diff
@@ -0,0 +1,369 @@
...
+If you don't want to wait until all the stages of the run are completed for the `pipeline` resource. You can provide the stage to be completed to trigger you pipeline. 
+```yaml
+resources:
+  pipelines:
+  - pipeline: SmartHotel
+    source: SmartHotel-CI 
+    trigger: 
+      branches:
+        include: master
+      stages:
```
ericsciple: Are the stages and'd or or'd

design/pipeline-triggers.md
ericsciple: I would imaging templates/tools to be the more common scenarios. I agree not trigger by default seems like the best behavior.

design/pipeline-triggers.md
ericsciple: Even for failed pipelines? I would default to succeeded or succeeded with issues.

ericsciple: (which matches the default condition on jobs/stages)

ericsciple: And how do you explicitly filter by result? Or are successful pipelines only supported for chaining?

design/pipeline-triggers.md
ericsciple: Note, spinnaker is a *continuous delivery* product. So that's not exactly an apples to apples comparison, since we're building a CI/CD pipeline product.

ashokirla: @vtbassmatt : I understand. Given the breaking changes, I think it should be fine if we can update the behavior in v1. As @chrisrpatterson mentioned we should improve the getting started experience.
@ericsciple : Yeah, I agree. But most of the scenarios that involve containers are CD scenarios. And our objective is to build a best in class Container CI/CD solution. So we intend do the right thing automatically.

design/pipeline-triggers.md
ashokirla: Okay, will remove the short syntax and will not mention anywhere in the spec.
Now, what if the user provides the short syntax since it is supported else where. Are we going to invalidate the YAML or does the parser have the support to understand this.

design/pipeline-triggers.md
ashokirla: The trigger section maps to the type. For Docker Registry, we don't have location. This property is available only if the type = ACR.

design/pipeline-triggers.md
ashokirla: Only the successful or partially succeeded pipelines are considered for triggers in this scope. We do the same in RM today. 

If we find valid scenarios for piplines with failed (jobs/stages), we can provide that support later on.

design/pipeline-triggers.md
vtbassmatt: It would be invalid YAML.

design/pipeline-triggers.md
ashokirla: Same as explained above.

design/pipeline-triggers.md
vtbassmatt: I may not fully understand the scenarios, so feel free to educate me. The behavior I'd naively expect is to OR everything. The "only if production && signed" case seems like an advanced scenario. Two ideas from me:

- Suggest a pattern of "put a step up front that checks for your custom conditions, halt the job if it doesn't match".
- Offer an alternate syntax that goes something like:
```yaml
...
  trigger:
    custom: >
      and(
        branchMatch(resources.foo.branch, 'releases/*'),
        eq(resources.foo.tag, 'Signed')
      )
```
(Obviously I just invented the `branchMatch` function; we would need to design the right set of functions here.)

vtbassmatt: Ha, silly me. Thanks Eric.

design/pipeline-triggers.md
ashokirla: Done changes as suggested.

design/pipeline-triggers.md
ashokirla: Discussed with @vtbassmatt. 
The inputs for tags/branches/stages are OR'd. 
For `AND` cases, we will introduce custom condition support inside trigger which takes expression input. This seems to be better and more powerful approach. This notation is already known to customers as part of step conditions.Later, when we introduce webhook triggers, this custom condition will be useful.

We need to define the functions appropriately. I will get back with the schema for this. However, since 'AND' is a corner case, we might pick this work in the next iteration.

design/pipeline-triggers.md
ashokirla: As discussed, removed the short syntax.

design/pipeline-triggers.md
ashokirla: Updated the scenario accordingly.

design/pipeline-triggers.md
ashokirla: As I have mentioned in the other conversation, for repos and containers, triggers will not be enabled by default in the current version of YAML.

design/pipeline-triggers.md
```diff
@@ -0,0 +1,359 @@
...
+
+You can control the target branches for your pull request based pipeline runs by simple syntax.
+```yaml
+resources:         
+  repositories:
+  - repository: myPHPApp      
+    type: GitHub
+    connection: myGitHubConnection
+    source: ashokirla/phpApp
+    pr:
```
ericsciple: where do pr fork settings live? (security problems in the file, which is why they live in the UI today)

ashokirla: I feel this is also dependent on how triggers are considered across branches. 

If we go with today's behavior of triggers in Build YAML, since triggers are considered only from the default branch used in the pipeline, the security problems will not creep in. Only those with access to the default branch can change the trigger settings.

ashokirla: As, discussed, we will take this incremental. Once we figure out the right place for pr fork policies for self-repos, we will replicate the same for other repositories.

PR #95, opened 2019-01-07
 Unify YAML template title/description format

DavidStaheli:
Deepak and Lova, could you please consider these changes that make the new YAML template titles/descriptions match the format of the existing built-in templates?  Thanks!


templates/docker-container-functionapp.yml
```diff
@@ -1,5 +1,5 @@
-# Docker image, Azure container registry and Function App
-# Build a Docker image, push to an Azure container registry and deploy to Azure Function App
+# Docker image, Azure Container Registry, and Azure Functions app
+# Build a Docker image, push to an Azure Container Registry, and deploy to an Azure Functions app.
```
ds-ms: "deploy `it` to an Azure Functions app" instead?

PR #102, opened 2019-01-15
Updating required inputs for templates

leantk:
Changing the templates' requires field to a list of required input objects. This will map to the TaskInputDefinition class on the server side. If you look at the actual yaml for these templates, the required inputs are serviceEndpointId and repositoryName. Basically azuresubscription (which is the current required input) is a serviceEndpointId that is of type connectedService:azureRM. On the server, azuresubscription  doesn't map to anything while the other name/type do. 

This additional information will allow us to take in these required inputs and replace them in the templates to get a complete yaml. The server side PR will be out after this one is completed.

### DIFF
--- a/templates/docker-container-functionapp.properties.json
+++ b/templates/docker-container-functionapp.properties.json
@@ -1,7 +1,7 @@
 {
   "iconName": "functionapp",
   "requires": [ 
-    { "name" : "serviceEndpointId", "type": "connectedService:azureRM" },
-    { "name" : "repositoryName", "type": "string" }
+    { "name" : "serviceEndpointId", "type": "connectedService:azureRM", "required": "true" },
+    { "name" : "repositoryName", "type": "string", "required": "true" }
   ]
 }
--- a/templates/docker-container-functionapp.properties.json
+++ b/templates/docker-container-functionapp.properties.json
@@ -1,6 +1,6 @@
 {
   "iconName": "functionapp",
-  "requires": [ 
+  "parameters": [ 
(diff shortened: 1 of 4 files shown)


templates/docker-container-functionapp.properties.json
```diff
@@ -1,4 +1,7 @@
 {
   "iconName": "functionapp",
-  "requires": [ "azuresubscription"]
+  "requires": [
```
hashtagchris: Given that TaskInputDefinition has a `Required` parameter, "parameters" would be a better property name than "requires". If you want to wait and make that change separately I think that's fine.

leantk: I'm just going to change it now since I was already thinking about it. I just added the extra parameter but I'll rename property as well. Just doing some testing

leantk: Ok I updated the server changes I'm making to adjust for this and everything still works 👍

templates/docker-container-functionapp.properties.json
```diff
@@ -1,4 +1,7 @@
 {
   "iconName": "functionapp",
-  "requires": [ "azuresubscription"]
+  "parameters": [ 
+    { "name" : "serviceEndpointId", "type": "connectedService:azureRM", "required": "true" },
+    { "name" : "repositoryName", "type": "string", "required": "true" }
```
ds-ms: Aren't all the parameters mentioned here, required?

PR #110, opened 2019-01-24
Create templates-2019.md

vtbassmatt:


mitchdenny:
Hi @vtbassmatt this feedback probably should go on the artifacts YAML but I'll add it here too. When we were looking at the simplified syntax for upload we think we encountered a problem:
```
- upload: **/*.exe
```
Is fine, but we need to assume what the base directory is for the minimatch. So I think we should default the root directory to $(System.DefaultWorkingDirectory). Does that work for you?

Then we have an optional additional property to specify the root. Something like:

```
- upload: **/*.exe
  root: build
```

vtbassmatt:
That seems fine.

vtbassmatt:
@mitchdenny I've thought this one through a bit more. If we're going to make a change to the `upload` schema, discussion needs to happen on that doc. This PR isn't the right place for deciding on new YAML syntax; it's only for describing updates to the templates.

PR #111, opened 2019-01-28
Use pytest-azurepipelines

vtbassmatt:
Instead of XML and a publish results step, use https://pypi.org/project/pytest-azurepipelines/

### DIFF
--- a/templates/python-package.yml
+++ b/templates/python-package.yml
@@ -33,15 +33,9 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
-      pip install pytest
-      pytest tests --doctest-modules --junitxml=junit/test-results.xml
+      pip install pytestpytest-azurepipelines
+      pytest tests --doctest-modules --test-run-title="Python $(python.version)"
     displayName: 'pytest'
-
-  - task: PublishTestResults@2
-    inputs:
-      testResultsFiles: '**/test-results.xml'
-      testRunTitle: 'Python $(python.version)'
-    condition: succeededOrFailed()
 
 - job: 'Publish'
   dependsOn: 'Test'
--- a/templates/python-package.yml
+++ b/templates/python-package.yml
@@ -33,7 +33,7 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
-      pip install pytestpytest-azurepipelines
+      pip install pytest pytest-azurepipelines
       pytest tests --doctest-modules --test-run-title="Python $(python.version)"
     displayName: 'pytest'
 


templates/python-package.yml
```diff
@@ -33,16 +33,10 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
-      pip install pytest
-      pytest tests --doctest-modules --junitxml=junit/test-results.xml
+      pip install pytestpytest-azurepipelines
```
brcrista: `pytest-azurepipelines` ?

brcrista:
I saw the plugin was released but haven't gotten a chance to play with it yet.  Do you have a test build with the new template?  Just want to make sure everything works with all the Python versions and platforms we support.

templates/python-package.yml
vtbassmatt: Python is truly a whitespace-significant language ;)

vtbassmatt:
I've not played too much. I found a project on GitHub which was mostly using the stock template. Verified that it [reports tests and things are passing](https://dev.azure.com/mattc-demo/Test1/_build/results?buildId=883), then [tried it again with the new tool](https://dev.azure.com/mattc-demo/Test1/_build/results?buildId=884). There are two new warnings about an escape sequence, so I'll dig into those.

brcrista:
Looks like the test runs are not getting named correctly?

Before:

image


After:

image

From the YAML, it looks like these should have the same name as before.

vtbassmatt:
I agree, and I can't quite figure out why it's not working. [This](https://dev.azure.com/mattc-demo/Test1/_build/results?buildId=886&view=logs) proves that the `$(python.version)` is still getting turned into the right value, so I can't figure out why the pytest-azurepipelines package sees a `$`.

vtbassmatt:
Oh, hmm, the `$` is a red herring. Even hardcoding a custom title doesn't seem to work.

hugovk:
Can this be closed now https://github.com/Microsoft/azure-pipelines-yaml/pull/141 is merged?

brcrista:
Ah, sorry @vtbassmatt , forgot about this PR.  I [tested this](https://dev.azure.com/brcrista/Python%20Brownbag/_build/results?buildId=727&view=ms.vss-test-web.build-test-results-tab) yesterday.  Looks like all the runs are named 'Pytest results' now: 
image

I had forgotten about this issue, but I do like how much the plugin simplifies the YAML.

vtbassmatt:
Yeah - looks like I have some more work to do to resolve the merge conflict.

vtbassmatt:
Whoops, Brian was saying he's already fixed the template. Sweet!

PR #121, opened 2019-02-09
design change to sources variables

vtbassmatt:
After reviewing in-box tasks and discussion with engineering, I don't like my earlier variables proposal. This one is less likely to break any assumptions, and moves us in a better direction for the future.

### DIFF
--- a/design/multicheckout.md
+++ b/design/multicheckout.md
@@ -49,17 +49,21 @@ It will be checked out into its default location, `scripts/` (driven by the reso
 
 ## Layout on disk and variables
 
-We introduce a default source directory: `$(Pipelines.DefaultSourcesDirectory)`.
-It's always `$(Agent.WorkDir)\s`.
-In single-checkout scenarios, the repo is still checked out directly in this directory.
-`$(Pipelines.SourcesDirectory)` (formerly known as `$(Build.SourcesDirectory)` also points here.
+`$(Build.SourcesDirectory)` (and someday, `$(Pipeline.SourcesDirectory)` always points to `$(Agent.WorkDir)\s`.
+In single-checkout, the repo is checked out directly there.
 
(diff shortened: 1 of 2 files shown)


design/multicheckout.md
```diff
@@ -49,17 +49,21 @@ It will be checked out into its default location, `scripts/` (driven by the reso
...
-In single-checkout scenarios, the repo is still checked out directly in this directory.
-`$(Pipelines.SourcesDirectory)` (formerly known as `$(Build.SourcesDirectory)` also points here.
+`$(Build.SourcesDirectory)` (and someday, `$(Pipeline.SourcesDirectory)` always points to `$(Agent.WorkDir)\s`.
+In single-checkout, the repo is checked out directly there.
 
-In multi-checkout scenarios, `$(Pipelines.DefaultSourcesDirectory)` serves as the default root for the multiple repos.
-`self` goes in `$(Pipelines.DefaultSourcesDirectory)\self`.
-A repository called `tools` goes in `$(Pipelines.DefaultSourcesDirectory)\tools` by default.
+In multi-checkout scenarios, it serves as the root for multiple repos.
+`self` goes in `$(Build.SourcesDirectory)\self`.
```
chrisrpatterson: Why would this not be ```$(Agent.WorkDir)\self```?

vtbassmatt: That's a breaking change. Tasks and scripts expect that source is rooted at `$(Agent.WorkDir)\s`. Also, I don't think we want to treat the `self` repo so specially going forward. It still gives you smart defaults in simple cases, but as you grow to have multi-repo and triggers and CD stages, it isn't the star of the show anymore.

vtbassmatt: We agreed in person to leave this under `$(Agent.WorkDir)\s`

PR #125, opened 2019-02-20
Reflect artifact changes discussed in working sessions

vtbassmatt:


vtbassmatt:
@TingluoHuang and @ericsciple -- would like to get your thoughts today so I can merge the PR and then we can get the work in quickly.

design/pipeline-artifacts.md
```diff
@@ -16,8 +18,8 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
 
 ```yaml
 - upload: string # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to upload
-  artifact: string # identifier for this artifact (no spaces allowed), defaults to 'default'
-  prependDirectory: string # a directory path that will be prepended to all uploaded files
+  artifact: string # identifier for this artifact (no spaces allowed), defaults to the job name
```
ericsciple: The job name is typically `__default`. Phase name is also problematic because it doesn't disambiguate the job configuration.

Consider something like: if job name is __default, use `<phase>`, otherwise use `<phase>.<job>`

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
 
-Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the directory specified in `root`. Each artifact is given its own directory e.g. `$PIPELINE_RESOURCESDIRECTORY\default` for the `default` artifact of this pipeline. Artifacts coming from other Pipelines are each given one directory per pipeline e.g. `$PIPELINE_RESOURCESDIRECTORY\some-other-pipeline\default` for the `default` artifact of the `some-other-pipeline` pipeline.
+#### YAML / unified pipelines
+Artifacts are downloaded to the pipeline's workspace by default. We introduce a new variable for unified pipelines: `$(Pipeline.Workspace)`. `Pipeline.Workspace` is one level up from `$(System.DefaultWorkingDirectory)`; on hosted, it corresponds to `c:\agent\_work\1\`.
+
+Each artifact is given its own directory e.g. `$(Pipeline.Workspace)\myartifact` for an artifact named `myartifact`. Artifacts coming from other pipelines are each given one directory per pipeline e.g. `$(Pipeline.Workspace)\some-other-pipeline\someartifact` for the `someartifact` artifact of the `some-other-pipeline` pipeline. The pipeline portion of the path comes from the ID of the pipeline's `resource` in YAML.
```
ericsciple: It sounds like this paragraph is about the implicit download behavior. Can you clarify?

The behavior you are describing seems good to me, for implicit download: don't add a `current` folder. But for other pipelines, do nest the artifacts in a folder named using the resource alias.

ericsciple: However, for explicit download, I would expect consistent behavior? (implicit is intentionally not consistent)

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
+No change to current behavior. Artifacts are downloaded to `$(System.DefaultWorkingDirectory)`, which is the sources folder on Build and the artifacts folder on RM.
+
+**Note**: this means we need a new job message type "Pipeline" in addition to the existing "Build" and "Release".
 
 ### Schema
 
 ```yaml
-- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline
-  artifact: string # identifier for the artifact to download; required
+- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline, blank means all available pipelines (including current)
```
ericsciple: does this translate to one task?

design/pipeline-artifacts.md
```diff
@@ -159,6 +245,26 @@ You can give an artifact a name, and you can upload multiple named artifacts. Al
...
-  - script: TODO-xamarin-magic $(Pipeline.ResourcesDirectory)/MobileApp/
-```
\ No newline at end of file
+  - script: ./my-deploy-script.sh $(Pipeline.Workspace)/WebApp/
+  - script: ./my-xamarin-script.sh $(Pipeline.Workspace)/MobileApp/
+```
+
+### Multiple artifact downloads with explicit path
+
+If a download step covers multiple artifacts and no explicit path is given, they're each pulled into a separate directory. With an explicit path, that explicit path becomes a containing folder for the artifacts.
```
ericsciple: Would it be simpler to explain and more predictable, to say that when an artifact name and path is specified, the artifact contents are downloaded to that folder. Otherwise artifacts are downloaded into a folder named using the artifact name.

Customers deserve a simple rule, for explicit control. And consistent behavior otherwise, that meets the mainline needs.

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
+
+**Note**: this means we need a new job message type "Pipeline" in addition to the existing "Build" and "Release".
 
 ### Schema
 
 ```yaml
-- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline
-  artifact: string # identifier for the artifact to download; required
+- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline, blank means all available pipelines (including current)
+  artifact: string # identifier for the artifact to download; optional
```
ericsciple: I'm not opposed to removing this restriction, but one download keyword should map to exactly one task instance.

We need to build and leverage the same infrastructure (as much as possible), that we think will enable third party extensibility in the future.

ericsciple: And the rules for how the directory structure is created, should be easy for customers to understand.

It would be good to explicitly write the details. And scrutinize whether the explanation is easy to understand.

Think through all cases. For example, explicit download for external pipeline, no artifact specified. A table might help think through the cases:

| Implicit/Explicit? | Pipeline has multiple artifacts? | Artifact specified? | Path specified? |
| --- | --- | --- | --- |
| Implicit | No | n/a | n/a |
| Implicit | Yes | n/a | n/a |
| Explicit | No | No | No |
| Explicit | No | No | Yes |
| Explicit | No | Yes | No |
| Explicit | No | Yes | Yes |
| Explicit | Yes | No | No |
| Explicit | Yes | No | Yes |
| Explicit | Yes | Yes | No |
| Explicit | Yes | Yes | Yes |

ericsciple: ...and whether the pipeline is `current` or an external pipeline should also be included in the matrix

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
 `download` is a shortcut for the [Download Pipeline Artifacts](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact) task.
 
-It will download artifacts uploaded from a previous job or stage or from another pipeline.
+It will download artifacts uploaded from a previous job, stage, or from another pipeline.
 
 ### Artifact download location
 
-Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the directory specified in `root`. Each artifact is given its own directory e.g. `$PIPELINE_RESOURCESDIRECTORY\default` for the `default` artifact of this pipeline. Artifacts coming from other Pipelines are each given one directory per pipeline e.g. `$PIPELINE_RESOURCESDIRECTORY\some-other-pipeline\default` for the `default` artifact of the `some-other-pipeline` pipeline.
+#### YAML / unified pipelines
+Artifacts are downloaded to the pipeline's workspace by default. We introduce a new variable for unified pipelines: `$(Pipeline.Workspace)`. `Pipeline.Workspace` is one level up from `$(System.DefaultWorkingDirectory)`; on hosted, it corresponds to `c:\agent\_work\1\`.
```
chrisrpatterson: On hosted this corresponds to D:\a\1

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
+  steps:
+  # by default, a download step is injected -- see later section
+  - script: ls $(Pipeline.Workspace)
+    # this listing shows a folder called `makeartifact`, since the prior job didn't specify a name
+
+- job: use2artifacts
+  dependsOn: makeartifact
+  steps:
+  - download: mypipe   # downloads all artifacts from `mypipe`
+  - download: current  # must include this, since by including a download step, we don't get automatic behavior anymore
```
chrisrpatterson: We need to make sure we reserve the alias current in the system.

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
+  - script: ./build.sh
+  - upload: outputs/**/*
+
+- job: useartifact
+  dependsOn: makeartifact
+  steps:
+  - download: current
+    path: foo
+  - script: ls $(Pipeline.Workspace)
+    # listing shows one folder, "foo"
```
chrisrpatterson: We should probably include an example directory tree that we are uploading and then show the resulting layout on disk after download just so it is clear given we are using a pattern match for the upload example.

design/pipeline-artifacts.md
```diff
@@ -37,23 +39,80 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
 ### Schema
 
 ```yaml
-- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline
-  artifact: string # identifier for the artifact to download; required
+- download: string # identifier for the pipeline resource from which to download artifacts, optional; "current" means the current pipeline, blank means all available pipelines (including current)
+  artifact: string # identifier for the artifact to download; optional
   patterns: string # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to download; if blank, the entire artifact is downloaded
-  root: string # the directory in which to download files, defaults to $PIPELINE_RESOURCESDIRECTORY; if a relative path is provided, it will be rooted from $SYSTEM_DEFAULTWORKINGDIRECTORY
+  path: string # the directory in which to download files, defaults to $(Pipeline.Worksapce); if a relative path is provided, it will be rooted from $(Pipeline.Workspace)
```
chrisrpatterson: I think we should start off by always requiring this to be a relative path.

design/pipeline-artifacts.md
```diff
@@ -83,6 +144,32 @@ Or to avoid downloading any of the artifacts at all:
 
 You can add files to any artifact multiple times in the same job, in different jobs in the same stage, and in different stages in the same pipeline, or you can `seal` an artifact at any time to prevent further files from being added.
 
+## Variables
+
+In addition to the new `$(Pipeline.Workspace)` variable, we introduce a variable per resource indicating where it was checked out on disk. For example:
```
ericsciple: What about explicit download, when the artifact name is specified. There is no \"resource\" directory. Are variables needed?

design/pipeline-artifacts.md
```diff
@@ -83,6 +144,32 @@ Or to avoid downloading any of the artifacts at all:
...
+  - script: ./build.sh
+  - upload: dist/**/*
+
+- job: consumer
+  dependsOn: maker
+  steps:
+  - download:   # remember, blank indicates we download all artifacts from all pipelines
+  - bash: |
+      echo $(Pipeline.Resources.foo)   # pipeline variable syntax
+      echo $PIPELINE_RESOURCES_FOO     # environment variable syntax
```
chrisrpatterson: Per our conversation earlier.  I wonder if we should choose to stop propagating everything into the environment variables and simply rely on either macro replacement in tasks or mapping using the ENV block for scripts.  There is a downside as it will make things a bit more complex in some cases for scripts run from the repo rather than inline.  There also could be a downside for task authors as there is now no way to get all of the artifact info from variables without the user explicitly giving you that input but I am not sure that is a requirement.

design/pipeline-artifacts.md
```diff
@@ -16,8 +18,8 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
 
 ```yaml
 - upload: string # a minimatch path or list of [minimatch paths](tasks/file-matching-patterns.md) to upload
-  artifact: string # identifier for this artifact (no spaces allowed), defaults to 'default'
-  prependDirectory: string # a directory path that will be prepended to all uploaded files
+  artifact: string # identifier for this artifact (no spaces allowed), defaults to the job name
+  prependPath: string # a directory path that will be prepended to all uploaded files
```
TingluoHuang: do we need both `prependPath` for upload and `path` for download?
What value `prependPath` can give us, since customer can use `path` to set download path to whatever they want?

chrisrpatterson: prepend path in the upload lets you move the set of files you are uploading into a specific hierarchy within the artifact without requiring you to move the files around on disk first.  This can be useful.

design/pipeline-artifacts.md
chrisrpatterson: I think variables are potentially needed and easier in the case you download to a specific path and just want to easy way to locate something. It has been pretty standard practice for our product for a while and it is the way we provide for information about objects in the pipeline for tasks.  I am not opposed to leaving off the environment variables as we move forward in a lot of these cases as it just bloats the environment and may not generally be necessary.  However, I think we will need to have pipeline variables that can be pulled from the context using either expressions or macros that cover the metadata about each of the resources.

design/pipeline-artifacts.md
vtbassmatt: Good call, thanks.

design/pipeline-artifacts.md
vtbassmatt: I intended this behavior in both cases (implicit, and explicit-with-no-path). Made that more clear.

design/pipeline-artifacts.md
vtbassmatt: Called that out explicitly.

design/pipeline-artifacts.md
vtbassmatt: Removed, since that was not adding much anyhow.

design/pipeline-artifacts.md
vtbassmatt: :+1:

design/pipeline-artifacts.md
vtbassmatt: Eric, was yours a technical concern or customer/usability concern? I agree with Chris this is a good thing for the customer to have, but if it's complicated to build then might not be worth it.

design/pipeline-artifacts.md
vtbassmatt: Yeah, we should make it work like `self`. I don't know where or if that's reserved.

design/pipeline-artifacts.md
vtbassmatt: This is really good feedback - thank you. I'll steal your table as a starting point.

design/pipeline-artifacts.md
mitchdenny: I think that having environment variables provides awesome flexibility for people to build their own stuff on top of what we provide. For example if someone wants to do some scanning on the artifacts before they deployment they can write a script which picks up all the artifact paths based on a prefix and feeds it into their scanning tools.

Is there a reason not to propagate in this particular case, it seems like it'll be helpful in some scenarios.

design/pipeline-artifacts.md
vtbassmatt: Simplifying the behavior.

design/pipeline-artifacts.md
vtbassmatt: Yeah

design/pipeline-artifacts.md
mitchdenny: Any reason why we wouldn't re-use ```self``` as the way to identify?

vtbassmatt: `self` refers to the repo. `current` refers to the pipeline run.

vtbassmatt: I don't remember all the discussion around this - but we had some concerns about making `self` ambiguous.

mitchdenny: OK - just making sure there was a reason to be different, sounds like there is.

design/pipeline-artifacts.md
```diff
@@ -35,30 +45,140 @@ Artifacts are distinct from other `resources` types, including `containers`, `re
...
 
 ### Artifact download location
 
-Artifacts are downloaded either to `$PIPELINE_RESOURCESDIRECTORY` or to the directory specified in `root`. Each artifact is given its own directory e.g. `$PIPELINE_RESOURCESDIRECTORY\default` for the `default` artifact of this pipeline. Artifacts coming from other Pipelines are each given one directory per pipeline e.g. `$PIPELINE_RESOURCESDIRECTORY\some-other-pipeline\default` for the `default` artifact of the `some-other-pipeline` pipeline.
+#### YAML / unified pipelines
 
-### Schema
+Artifacts are automatically downloaded to the pipeline's workspace by default. We introduce a new variable for unified pipelines: `$(Pipeline.Workspace)`. `Pipeline.Workspace` is one level up from `$(System.DefaultWorkingDirectory)`. This is the "shallowest" place where artifacts may be placed.
+
+The spec that follows is lengthy because it covers details and edge cases.
```
vtbassmatt: @ericsciple in lieu of a table, I boiled this down to its 3-bullet-point essence. I think you'll approve.

PR #126, opened 2019-02-21
Fixing parameter names in properties.json files

ds-ms:
The properties.json mentions it requires a parameter called `serviceEndpointId` but the yaml uses `azureServiceConnectionId`.
This PR fixes that.


atbagga:
Thanks for fixing this. 
We are taking a dependency on these names so the I hope they will not change too soon now. :)

PR #128, opened 2019-02-22
Step target

chrispat:


vtbassmatt:
This makes sense to me.

design/step-target.md
```diff
@@ -0,0 +1,134 @@
+# Step Target
+
+Steps run either on the host that is running the agent.worker or if the step is part of a [Container Job](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops&tabs=yaml) it runs inside of the container specific on the job.  Steps are not able to target the host if running in a container job or other containers if using [services](./sidecar-containers.md).
```
stephenmichaelf: specific on the job -> specified on the job?

TingluoHuang:
does this means we have to do the same setup for all the service container? ex: create user/group and make sure it can run node.

chrisrpatterson:
@TingluoHuang maybe but I am not as concerned there.  Perhaps we should think about other options.

vtbassmatt:
Should the reserved word for targeting the host be `host` or something else? I generally like `host` except for cases where the agent itself is running in a container - then it's ambiguous. `none` might work, or `agent`. @jtpetty any thoughts?

jtpetty:
I'm ok with ```host```. Even in the case where the agent is running inside a container, from the point of view of the agent, it is the host. ```parent```, ```main```, ```outer``` could all work. I don't like ```agent``` since we are spawn separate processes and we do not change the state of the agent with the step. I do not like ```none``` as an value for target, as that seems confusing. If the key was ```container``` or ```target_container```, I think ```none``` could work.

vtbassmatt:
I'm good with `host`, since as you point out, from the agent's point of view, its container is its host.

PR #129, opened 2019-02-22
Runtime parameters

vtbassmatt:
This needs fleshing out, but I wanted to put it up early for your thoughts.


design/process-parameters.md
```diff
@@ -0,0 +1,65 @@
...
+- The pipeline itself can be altered at queue time: pools, agentspecs, service connections, environments, and more
+
+## Schema
+
+_Note: we need to rationalize this with the existing `parameters` syntax._
+
+```yaml
+parameters:
+# this is a sequence, so the block may be repeated as many times as needed
+- name: string          # name of the parameter; required
```
chrisrpatterson: Why would we make this significantly different than the schema we have for Task inputs?  It seems the general constraints are the same.

vtbassmatt: Great question. I used the task schema as guidance. I felt like we've learned a lot since then, plus, the expected number of YAML authors at least is an order of magnitude higher than the number of task authors. Where it made sense, I shortened things or matched YAML precedent, since the bulk of users will not be familiar with task.json.

- **name**: matches
- **displayName**: matched our YAML model instead of task.json's _label_
- **type** / **displayHint**: this is the major departure, and its reasoning is discussed lower in the spec. "Radio" and "pickList" aren't types, they're UI widgets, and I wanted to separate them from the data type.
- **default**: in task.json it's defaultValue; seemed verbose 
- **range**: the precedent in task.json is data binding. By making it an expression-ish field, we can add that later but start by supporting simple things like enums and integer ranges.

Missing from YAML, present in task.json:
- _required_: largely meaningless in task.json, so I didn't bring it forward
- _groupName_ and _visibleRule_: didn't want to bring forward this complexity, we can add it later if required
- _properties_ and _options_: unclear what these would be needed for in YAML

Of the above, the interesting ones to discuss are **range** and **type**/**displayHint**. I'd like to get your thoughts on those specifically.

vtbassmatt: (need to add a value for "allowed at queue time")

vtbassmatt: (add "object" as a type - also consider if "job"/"step"/etc should be well-known)

vtbassmatt: `range` is a little weird, find a different way

design/process-parameters.md
```diff
@@ -0,0 +1,230 @@
...
+parameters:
+- name: 'image'     # scalar parameter
+  type: string
+  default: 'ubuntu-16.04'
+- name: 'preSteps'  # sequence parameter
+  type: object
+  default:
+  - script: echo foo
+- name: 'vars'      # map parameter
+  type: object
```
chrisrpatterson: Do we ever let a customer attach a schema here to define what the object should look like?

vtbassmatt: Sounds like a good future enhancement

design/runtime-parameters.md
```diff
@@ -0,0 +1,226 @@
...
+# azure-pipelines.yml - in the Roslyn repo
+
+resources:
+  repositories:
+  - repository: arcade
+    name: dotnet/arcade
+
+steps:
+- template: template.yml@arcade
+  parameters:
```
jeremyepling: I'm not following this example. What happens here?

vtbassmatt: The Arcade team offers a template which expects a list of steps. The Roslyn team uses that template and passes in their list of steps.

Today, it's completely untyped. Arcade has to blindly assume that Roslyn (and other customers) gave valid YAML for steps. It will fail with mysterious errors if they don't. With this change, we will know in the parser that Arcade expected a list of steps, and if the YAML isn't right, we can give a better error. It also opens the door to offering editor support: when the consumer uses a template that expects a list of steps, we'll be able to IntelliSense and syntax-highlight them correctly.

jeremyepling: Ah, I read this set of parameters as run parameters not template parameters.

design/runtime-parameters.md
```diff
@@ -0,0 +1,226 @@
...
+| `filePath`
+| `secureFile`
+| `pool`
+| `serviceConnection`
+| `environment`
+
+It's also useful to tighten up the editor experience and error messages for template parameters.
+Customers are already using templates to pass along step and job lists.
+These aren't very useful for run-time parameters: it's unclear how a user would be expected to set them.
+They're rendered like a plain `object` in the UI.
```
jeremyepling: What is the change you're making?

jeremyepling:
A very common scenario will be CI+CD pipelines that just want to run CI or CD, or a specific stage or deployment. Can you add examples for that? I'm especially interested in the UX design for these. Do have "built in" support for running a single stage or deployment? I'd like to have that so customer don't need to add a lot of variables. What are the most common scenarios?

@vijayma @RoopeshNair @azooinmyluggage

design/runtime-parameters.md
vtbassmatt: More confusion between runtime and template parameters. Fixing.

TroySchmidt:
Can this be given a higher priority?  Not being able to set variables at run time for multi stage YAML builds kinda makes the ability mute.  Currently I bet a lot of people have copied builds with different steps and parameters setup because there is little way to control what steps run.  Parameters that can be overridden at runtime is the only way to truly allow single multistage YAML builds to handle everything.

vijayma:
Thanks for the feedback. I would like to point out that you can define variables for your YAML pipeline in the UI and mark them as "settable at queue time". I agree that it would be nice to be able to do this from within the YAML file, and we are starting the work for this.

TroySchmidt:
@vijayma While it is true that you can use the UI to create and set those they **only** work for conditions when you translate those back into variables and then evaluate in conditions.  That workaround also **only** works at the job \ task level.  When using multi stage YAML definitions it is tedious to copy and paste each condition evaluation to every single task.  

Otherwise expansion evaluation of the variables doesn't work at any higher level.

```YAML
- job
  variables:
    evalVar: $(setAtRuntimeVar)
  steps:
  - task: ANY
    condition: eq(variables.evalVar, True)
```

design/runtime-parameters.md
```diff
@@ -0,0 +1,339 @@
...
+  type: number
+  default: 123
+- name: mypool
+  displayName: 'Pool'
+  type: pool
+  values:
+  - ubuntu-latest
+  - windows-latest
+  - macos-latest
+
```
jagany: In classic ADO build or release we go all with Variables and Variable groups. Here in YAML we have something like parameters and also variables. During build save or queue time does it show up parameters as variables or variables itself? What about global variables and system defined variables to display at queue time for user to change it if needed. Ex: System.Debug = true or false.

PR #133, opened 2019-02-28
Deployment strategies

RoopeshNair:
Support for deployment strategies in deployment job


design/deployment-strategies.md
```diff
@@ -0,0 +1,212 @@
...
+      routeTraffic:
+        delay: 60m
+        steps:
+        - task: AzureAppServiceManage@0
+          inputs:
+            Action: 'Swap Slots'                                    # swap green with blue
+            WebAppName: 'musicCarnivalWeb'
+            sourceSlot: $(environment.staging)
+            destinationSlot: $(environment.prod)
+      postRouteTafficCheck:
```
chperich: Can we just call this postRouteTaffic instead?

design/deployment-strategies.md
```diff
@@ -0,0 +1,212 @@
...
+  strategy:                 
+    runOnce:              
+      deploy:    
+        steps:             
+        - script: echo deploy web app...   
+ ```
+
+When deploying application updates it is important that the technique used to deliver update enables initialization, deploying the update, testing the updated version after routing traffic and in case of failure, run steps to restore to last known good version. We achieve this using fundamental building blocks viz., hooks where you can run your steps during deployment lifecycle. Following are descriptions of the lifecycle events where you can run a hook during a deployment lifecycle
+
+**preDeploy** – Use to run tasks before the deploy step is executed. A rollback is not possible at this point.
```
chperich: This statement needs to be corrected. We need to run onFailure even for failures in preDeploy step.

design/deployment-strategies.md
```diff
@@ -0,0 +1,212 @@
...
+Consider the following Azure Pipelines pipeline defined in YAML
+
+```yaml
+jobs:
+- deployment:
+  environment: musicCarnivalProd
+  pool:
+    name: musicCarnivalProdPool 
+  strategy:                 
+    canary:
```
chperich: Can you please specify how each of the increments input will map to a new iteration and the list of variables that will be available for the tasks in the iteration.

design/deployment-strategies.md
RoopeshNair: Good point, user can run any step after routing traffic not just checks. The pool type would determine, if it's agent or server or container

design/deployment-strategies.md
RoopeshNair: Each iteration will include the following variables $(increments), $(strategy), and $(action) in case of canary.  These variables can be used/mapped to the step inputs in case of k8s-deploy

PR #136, opened 2019-03-04
Remove the trigger on 'master' in favor of [skip ci] on wizard commits

DavidStaheli:
The "New pipeline" wizard will now append `[skip ci]` to commits so that a **manually**-queued build still runs for a new pipeline, but a **CI** build will not.  With this, we can remove the explicit trigger on 'master' from the YAML templates since `[skip ci]` will have the desired effect.

### DIFF
--- a/templates/.net-desktop.yml
+++ b/templates/.net-desktop.yml
@@ -3,9 +3,6 @@
 # Add steps that publish symbols, save build artifacts, and more:
 # https://docs.microsoft.com/azure/devops/pipelines/apps/windows/dot-net
 
-trigger:
-- master
-
 pool:
   vmImage: 'VS2017-Win2016'
 
--- a/templates/android.yml
+++ b/templates/android.yml
@@ -3,9 +3,6 @@
 # Add steps that test, sign, and distribute the APK, save build artifacts, and more:
 # https://docs.microsoft.com/azure/devops/pipelines/languages/android
 
-trigger:
-- master
-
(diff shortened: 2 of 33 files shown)


jayspadie:
So we've added [skip ci] to commits created by the YAML editor, but that's only part of the story for specifying only the trigger for master.  Removing the trigger will reintroduce CI builds for commits to every branch.  Do we really want to change that behavior?

DavidStaheli:
Hmm, good point.  I think it will cause a CI build even for PR branches with commits that don't use [skip ci].

jayspadie:
> Hmm, good point. I think it will cause a CI build even for PR branches with commits that don't use [skip ci].

Right... This goes beyond whatever our YAML editor does.  It changes the CI behavior for all commits in all branches (with or without a PR), triggering a build for each push.  Maybe some customers want this, but it's a big behavior change by default.

madhurig:
We switched from not having the trigger to adding it and we didn't get any push back on that. I personally like having the trigger in the yml, it is very explicit and clear to the user why builds are triggered only on master branch.

jayspadie:
Yes, to be clear, I like the trigger as well and I don't think we should remove it.  It gives the user clear awareness of the trigger option, it prevents the proliferation of unwanted builds, and the user can update it as they see fit.

DavidStaheli:
Ba-bam.  Thanks for the feedback on this!

PR #141, opened 2019-03-08
Use Pytest plugin for Azure Pipelines

brcrista:
The plugin automatically produces and uploads JUnit XML, among some other great things: https://pypi.org/project/pytest-azurepipelines/

Includes some other small simplifications to the template.


brcrista:
@tonybaloney this is the YAML that gets suggested when a user creates a pipeline for a Python repo.  Great job on the plugin!

PR #143, opened 2019-03-12
Updating AKS template to contain information about service end point created

rakol-ms:
W.R.T recent work in configless CD we are creating service end points and are integrating the K8S resource with other flows. Adding the parameters in template so that the created information can be pulled.


templates/docker-container-to-aks.properties.json
```diff
@@ -6,6 +6,8 @@
     { "name" : "kubernetesConnectionId", "type": "create:azureKubernetesEndpoint", "required": "true"},
```
ds-ms: Rename this to `kubernetesConnection` instead, as you're accessing it as an object.

PR #148, opened 2019-03-19
Excluding path support for Q1

ashokirla:
Due to few dependencies we are planning to move path support in download pipelines shortcut to post Q1 timeline.


vtbassmatt:
I don't think we should delete the things we want to do in the future. Instead, we should mark them as "future" or "p2" or something.

PR #151, opened 2019-03-24
fixing the download task example

kasubram:
Fixing the download task example to use the artifact keyword


ashokirla:
Fixed few fit and finish changes to downloadBuild macro and build resources.

PR #152, opened 2019-03-28
Deploy to kubernetes

ds-ms:
This PR adds a new template which builds a docker image and deploys on the k8s cluster.
It utilizes the newly released tasks i.e. `Docker@2` and `KubernetesManifest@0`


templates/deploy-to-k8s.yml
```diff
@@ -0,0 +1,79 @@
...
+  dependsOn: Build
+  jobs:
+  - deployment: Deploy to Kubernetes
+    displayName: Deploy job
+    environment: $(Environment.Name).$(Environment.k8sResource.Name)
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+          - task: Kubernetes@2
```
shashankbarsin: This needs to be replaced with createSecret action of KubernetesManifest task

templates/deploy-to-k8s.yml
```diff
@@ -0,0 +1,79 @@
...
+              secretName: 'docker-auth'
+              
+          - task: KubernetesManifest@0
+            displayName: Deploy to Kubernetes cluster
+            inputs:
+              action: deploy
+              manifests: |
+                manifests/*
+              containers: |
+                $(ContainerRegistry.Name)/$(Image.Repository)
```
shashankbarsin: secrets input needs to be added to the task with the name of the docker registry secret created in preceding action. This is so that the manifest file can be augmented with imagePullSecrets before the files are applied

templates/deploy-to-k8s.properties.json
```diff
@@ -0,0 +1,31 @@
...
+            "displayName": "Image Name"
+        },
+        {
+            "name": "servicePort",
+            "type": "string",
+            "required": "true",
+            "displayName": "Service Port"
+        }
+    ],
+    "fileAssets": [
```
ds-ms: We can either have it like this. 
Make it like 
```json
"assets": [{
  "type": "file", 
  "path": "<filePath>"
}]
```

templates/deploy-to-k8s.yml
ds-ms: Once the changes are done to the task, yes.

templates/deploy-to-k8s.yml
ds-ms: Yes. Once the task is updated. the task won't be able to make sense out of `secrets` right now.

templates/deploy-to-k8s.properties.json
```diff
@@ -0,0 +1,37 @@
+{
+    "iconName": "k8s",
```
bansalaseem: Should not the name be deploy-to-existing-k8s.properties.json

templates/deploy-to-k8s.yml
```diff
@@ -0,0 +1,72 @@
...
+trigger:
+- master
+
+resources:
+- repo: self
+
+variables:
+
+  # Container registry service connection established during pipeline creation - Docker Hub/ACR/...
+  DockerRegistry.ServiceConnection: ''
```
bansalaseem: Please rename the variables to lower case.

templates/deploy-to-k8s.yml
```diff
@@ -0,0 +1,72 @@
...
+          - task: Kubernetes@2
+            displayName: 'kubectl create imagepullsecret docker-auth'
+            inputs:
+              containerRegistryType: 'Container Registry'
+              dockerRegistryEndpoint: $(DockerRegistry.ServiceConnection)
+              secretName: 'docker-auth'
+              
+          - task: KubernetesManifest@0
+            displayName: Deploy to Kubernetes cluster
+            inputs:
```
shashankbarsin: namespace input needs to be set. Set its value to $(Environment.k8sResource.Name)

PR #155, opened 2019-04-01
Modifying template for docker registry

rakol-ms:


templates/deploy-to-existing-k8s.yml
```diff
@@ -11,9 +11,9 @@ resources:
 variables:
 
   # Container registry service connection established during pipeline creation - Docker Hub/ACR/...
-  dockerRegistryServiceConnection: ''
-  imageRepository: ''
-  containerRegistryName: ''
+  dockerRegistryServiceConnection: '{{ containerRegistryConnection }}'
+  imageRepository: '{{ containerRegistryConnection }}'
+  containerRegistryName: '{{ containerRegistryConnection }}'
```
ds-ms: Shouldn't it be an endpoint object here? And use it like `containerRegistryConnection.Id`

PR #156, opened 2019-04-01
Adding upload: statement to templates.

mitchdenny:
This is an initial pass at getting the ```upload:``` statement into the YAML templates. For discussion at this stage and we'll merge it once any issues are ironed out. I've updated almost all templates with a single ```- upload:``` statement that uploads the entire working directory (sans the ```.git``` folder). The artifact name will be automatically generated based on the job name (this work is in progress, and if we don't get it in time we'll need to hardcode the artifact name to avoid breaking builds when we do implement automatic artifact naming.

### DIFF
--- a/templates/android.yml
+++ b/templates/android.yml
@@ -18,3 +18,5 @@ steps:
     publishJUnitResults: false
     testResultsFiles: '**/TEST-*.xml'
     tasks: 'assembleDebug'
+
+- upload:
--- a/templates/.net-desktop.yml
+++ b/templates/.net-desktop.yml
@@ -31,3 +31,5 @@ steps:
   inputs:
     platform: '$(buildPlatform)'
     configuration: '$(buildConfiguration)'
+
+- upload: '**/bin/{$buildConfiguration}'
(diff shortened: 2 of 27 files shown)


templates/android.yml
```diff
@@ -18,3 +18,5 @@ steps:
     publishJUnitResults: false
     testResultsFiles: '**/TEST-*.xml'
     tasks: 'assembleDebug'
+
+- upload:
```
vtbassmatt: Do we want a bunch of content-less `upload:` statements? I know it works, but would it be nicer to put a minimatch of `**/*` here so that it's obvious to users what's getting uploaded?

mitchdenny: We don't have filtered upload yet so ```**/*``` is misleading. But we could do ```- upload: $(System.DefaultWorkingDirectory)``` but that looked pretty ugly. On filtered upload - a minimatch pattern makes sense if you know what the root directory is (implied is the defautl working directory).

I'm going to start a side thread on this because I've been wrestling with this for a while so I want to get some input to help get my thinking straight.

PR #158, opened 2019-04-04
[Deploy to K8s] Clean up

ds-ms:
In this PR
1. Unused variables are removed
2. Using single quotes everywhere instead of double in this template
3. Updated the task usage to sync with the task changes here https://github.com/Microsoft/azure-pipelines-tasks/pull/10029


templates/deploy-to-existing-k8s.yml
```diff
@@ -12,19 +12,12 @@ variables:
...
-  dockerfilePath: '**/Dockerfile'
-
-  # Kubernetes Environment
-  environmentName: '{{ environment.Environment.Name }}'
-  k8sResourceName: "{{ environment.Resources.kubernetesConnection.Name }}"
-  k8sResourceNamespace: "{{ environment.Resources.kubernetesConnection.Namespace }}"
-
-    # Docker Image Name
   imageName: '{{ imageName }}'
+  containerRegistryName: '{{ containerRegistryName }}.azurecr.io'
```
shashankbarsin: Why this change? This way we'll have to change the template again when we add support in UX for choosing container registry other than ACR

templates/deploy-to-existing-k8s.yml
```diff
@@ -47,24 +40,24 @@ stages:
...
-    environment: $(environmentName).$(k8sResourceName)
+    environment: '{{ environment.Environment.Name }}.{{ environment.Resources.kubernetesConnection.Name }}'
     strategy:
       runOnce:
         deploy:
           steps:
-          - task: Kubernetes@2
-            displayName: 'kubectl create imagepullsecret docker-auth'
+          - task: KubernetesManifest@0
+            displayName: kubectl create imagepullsecret docker-auth
```
shashankbarsin: Change to ```displayName: Create imagePullSecret```

templates/deploy-to-existing-k8s.yml
```diff
@@ -12,19 +12,12 @@ variables:
...
-
-  dockerfilePath: '**/Dockerfile'
-
-  # Kubernetes Environment
-  environmentName: '{{ environment.Environment.Name }}'
-  k8sResourceName: "{{ environment.Resources.kubernetesConnection.Name }}"
-  k8sResourceNamespace: "{{ environment.Resources.kubernetesConnection.Namespace }}"
-
-    # Docker Image Name
   imageName: '{{ imageName }}'
```
shashankbarsin: Change this back to imageRepository. Let's stick to Docker registry nomenclature

templates/deploy-to-existing-k8s.yml
```diff
@@ -47,24 +40,24 @@ stages:
...
             displayName: Deploy to Kubernetes cluster
             inputs:
               action: deploy
-              namespace: $(k8sResourceNamespace)
+              namespace: $(k8sNamespace)
               manifests: |
                 manifests/*
               containers: |
-                $(containerRegistryName)/$(imageRepository)
+                $(containerRegistryName)/$(imageName)
```
shashankbarsin: Change back to $(imageRepository). Introduce one more value at the top section called tag and set its value to $(Build.BuildId). 

While this is already the default in Docker task, still set it explicitly in Docker task so that it is clear to the users as most folks using templates may be new to Azure Pipelines YAML syntax. 

After that append  $(tag) to $(containerRegistryName)/$(imageRepository) in KubernetesManifest task

PR #163, opened 2019-04-05
Ashokirla resources updates

ashokirla:
I have added few clarifications to build resource spec and updated the pipeline artifacts spec regarding supporting `download` macro with the scenario - download all available artifacts to a specified path.

### DIFF
--- a/design/pipeline-artifacts.md
+++ b/design/pipeline-artifacts.md
@@ -143,7 +143,7 @@ However, if you include a `path` but don't mention any `artifact`, all the avail
 Example:
 ```yaml
 jobs:
-- job: myjob
+- job: job1
   steps:
   - script: ./build.sh
   - upload: outputs/**/*
@@ -151,8 +151,8 @@ jobs:
   - upload: daata/**/*
     artifact: DBTier
 
-- job: useartifact
-  dependsOn: makeartifact
+- job: job2
+  dependsOn: job1
   steps:
   - download: current
     path: foo
--- a/design/pipeline-artifacts.md
+++ b/design/pipeline-artifacts.md
@@ -89,7 +89,7 @@ The spec that follows is lengthy because it covers details and edge cases.
 But for customers, it should be easy to understand:
 1. If you don't say anything, you get a directory per artifact from the current pipeline. You also get a directory per pipeline containing a directory per artifact from other pipelines.
 2. As soon as you put a `download` step, you have to be explicit about all pipelines you want to download from. You still get  automatic directory layout.
-3. You may mention a specific artifact, but you don't have to unless you mention a `path`. As soon as you mention a `path`, you _must_ also choose the artifact. This is the mode with ultimate control/flexibility.
+3. You may mention a specific artifact, in which case only the mentioned artifact is downloaded to the automatic directory layout. If you also mention a `path`, the contents of the artifact are downloaded directly to the path provided.
 
 ##### With no `path`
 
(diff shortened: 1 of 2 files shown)


design/pipeline-artifacts.md
```diff
@@ -59,19 +59,25 @@ Listing just the unique properties of the `download` step:
...
 
-The "magic" of download is how often you _don't_ have to specify it.
-By default, all artifacts produced in the current pipeline and all artifacts in all referenced pipelines will be downloaded.
-As soon as you include an explicit `download` step (or task), the user is in total control - no more automatic download for that job.
-Similarly, you can leave out `artifact`, `patterns`, or `path` to get some automatic behaviors (described below).
+Usually a 'job' publishes build artifacts which are deployed to an environment as part of the 'deployment' job.
+Therefore, the automatic artifact download behavior varies between a 'job' and 'deployment'.
+
+#### job
+Artifacts are not automatically downloaded in regular jobs. However, if you require artifacts to be downloaded in your jobs, you need to explicitly use `download`.
```
willsmythe: Just to be clear, I would reword the second sentence: "To download artifacts in these jobs, use the `download` task.`

design/pipeline-artifacts.md
```diff
@@ -89,7 +95,7 @@ The spec that follows is lengthy because it covers details and edge cases.
 But for customers, it should be easy to understand:
 1. If you don't say anything, you get a directory per artifact from the current pipeline. You also get a directory per pipeline containing a directory per artifact from other pipelines.
 2. As soon as you put a `download` step, you have to be explicit about all pipelines you want to download from. You still get  automatic directory layout.
-3. You may mention a specific artifact, but you don't have to unless you mention a `path`. As soon as you mention a `path`, you _must_ also choose the artifact. This is the mode with ultimate control/flexibility.
+3. You may mention a specific artifact, in which case only the mentioned artifact is downloaded to the automatic directory layout. If you also mention a `path`, the contents of the artifact are downloaded directly to the path provided.
```
willsmythe: Good change here ... @vtbassmatt and I were just discussing this scenario and didn't think the originally-proposed behavior made sense / would be expected by users.

design/pipeline-artifacts.md
```diff
@@ -131,16 +137,19 @@ jobs:
...
+- job: myjob
   steps:
   - script: ./build.sh
   - upload: outputs/**/*
+    artifact: WebTier
+  - upload: daata/**/*
+    artifact: DBTier
 
 - job: useartifact
   dependsOn: makeartifact
```
willsmythe: `dependsOn` should be `myjob`. Additionally, it might be clearer to just call the first job `job1` and the second `job2`.

design/pipeline-artifacts.md
```diff
@@ -150,6 +159,7 @@ jobs:
   - script: ls $(Pipeline.Workspace)
     # listing shows one folder, "foo"
 ```
+In this case the artifact contents for each artifact is downloaded to `$(Pipeline.Workspace)/foo/WebTier/` and `$(Pipeline.Workspace)/foo/DBTier/` respectively.
```
willsmythe: where does 'foo` come from?

design/pipeline-resources.md
```diff
@@ -83,7 +83,7 @@ Or to avoid downloading any of the artifacts at all:
 - download: none
 ```
 
-Artifacts from the `pipeline` resource are downloaded to `$PIPELINES_RESOURCESDIR/<pipeline-identifier>/<artifact-identifier>` folder; see [artifact download location](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md#artifact-download-location) for more details.
+Artifacts from the `pipeline` resource are downloaded to `$(PIPELINE.WORKSPACE)/<pipeline-identifier>/<artifact-identifier>` folder; see [artifact download location](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/pipeline-artifacts.md#artifact-download-location) for more details.
```
willsmythe: what if there are folder name conflicts? for example, the current pipeline has an artifact named `foo`, and there is a `foo` pipeline resource ...

design/pipeline-artifacts.md
ashokirla: Updated as suggested.

design/pipeline-artifacts.md
ashokirla: Here 'foo' is the relative path into which the artifacts are downloaded.

design/pipeline-resources.md
ashokirla: The name of the artifact published by the current pipeline is set in the YAML. Pipeline identifier is a unique alias for the pipeline resource user provides in the YAML. User needs to ensure the pipeline identifier is unique to avoid collisions. These are very corner scenarios and can be avoided by the user during the pipeline YAML authoring time. In future we can provide queue time validations and warn users of possibility for collisions. 
We want to provide a simpler experience for CICD and CD scenarios.

design/pipeline-artifacts.md
ashokirla: Good catch. Updated it now.

ashokirla:
Addressed all the comments. Merging this PR.

PR #164, opened 2019-04-05
YAML templates refer to the 'latest' hosted OS version

DavidStaheli:
This updates the YAML templates to refer to the 'latest' available OS version on Microsoft-hosted images, instead of pinning them to a version that ages.  These templates will be ingested into the Azure DevOps codebase and deploy with sprint 151.

### DIFF
--- a/templates/.net-desktop.yml
+++ b/templates/.net-desktop.yml
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
-  vmImage: 'VS2017-Win2016'
+  vmImage: 'Windows-latest'
 
 variables:
   solution: '**/*.sln'
--- a/templates/.net-desktop.yml
+++ b/templates/.net-desktop.yml
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
-  vmImage: 'Windows-latest'
+  vmImage: 'windows-latest'
 
 variables:
   solution: '**/*.sln'
--- a/templates/android.yml
+++ b/templates/android.yml
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
-  vmImage: 'macOS-latest'
+  vmImage: 'macos-latest'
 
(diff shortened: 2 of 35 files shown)


templates/.net-desktop.yml
```diff
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
-  vmImage: 'VS2017-Win2016'
+  vmImage: 'Windows-latest'
```
vtbassmatt: Aesthetically, I don't love the combination of capitalizing "Windows" but including the hyphen. It's a nit, though.

templates/android.yml
```diff
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
-  vmImage: 'macOS-10.13'
+  vmImage: 'macOS-latest'
```
vtbassmatt: Same for Mac and Linux... wish they were all lowercase, but it's a nit

madhurig: +1 for all lower case.

fowl2:
Given that using `latest` is generally considered a Bad Idea^, is using them in templates - which many use a example of good practices - really a good one?

^ as it means you can't actually tell what image is being used, and it will change over time

DavidStaheli:
Hi @fowl2.  Thank you for commenting on this because we certainly want to do the right thing and not disrupt people.  Here is some of the thinking behind this:

1. Soon there will be image information in the logs so that it's clearer which versions are being used by each run of a pipeline.
2. We plan to add update notifications so that people know when versions will change.
3. These 'latest' image names are to address what we hear from most people - that they want to move with the latest OS versions as they get released.  As you point out, that can also break people's pipelines.

For people who don't want to pin to the _latest_ OS release, we may be able to make it clearer in the templates how to pin to a _specific_ version.

Your thoughts and ideas are much appreciated.

PR #165, opened 2019-04-08
Moving container registry to be picked from auth data

rakol-ms:


templates/deploy-to-existing-kubernetes-cluster.yml
```diff
@@ -13,7 +13,7 @@ variables:
   # Container registry service connection established during pipeline creation - Docker Hub/ACR/...
   dockerRegistryServiceConnection: '{{ containerRegistryConnection.Id }}'
   imageRepository: '{{ imageRepository }}'
-  containerRegistryName: '{{ containerRegistryName }}'
+  containerRegistry: '{{ containerRegistryConnection.Authorization.Parameters.loginServer }}'
```
bansalaseem: can you please remove the parameter from props file in that case?

ds-ms: This is not added to properties.json file

PR #171, opened 2019-04-16
Adding template for nodejs function app on linux, and modifying Express and React templates to have separate Build and Deploy stages.

ajinkya599:
- Adding template to build and deploy node.js function app on Linux.
- Node.js Express web app - separate stages for Build and Deploy
- Node.js React web app - separate stages for Build and Deploy

### DIFF
--- a/templates/node.js-express-webapp-on-linux.yml
+++ b/templates/node.js-express-webapp-on-linux.yml
@@ -57,23 +57,26 @@ stages:
   dependsOn: Build
   condition: succeeded()
   jobs:
-  - job: Deploy
+  - deployment: Deploy
     displayName: Deploy job
-    pool:
-      vmImage: $(vmImageName)
-    
-    steps:
-    - task: DownloadPipelineArtifact@1
-      displayName: 'Download Pipeline Artifact'
-      inputs:
-        buildType: 'current'
-      
-    - task: AzureWebApp@1
-      displayName: 'Azure Web App Deploy: $(webAppName)'
-      inputs:
-        azureSubscription: $(azureSubscription)
-        appType: webAppLinux
-        appName: $(webAppName)
-        runtimeStack: 'NODE|10.10'
-        package: $(System.ArtifactsDirectory)/drop/$(Build.BuildId).zip
-        startUpCommand: 'npm run start'
+    environment: 'development'
+    pool: 
+      vmImage: $(vmImageName)    
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+          - task: DownloadPipelineArtifact@1
+            displayName: 'Download Pipeline Artifact'
+            inputs:
+              buildType: 'current'
+            
+          - task: AzureWebApp@1
+            displayName: 'Azure Web App Deploy: {{ webAppName }}'
+            inputs:
+              azureSubscription: $(azureSubscription)
+              appType: webAppLinux
+              appName: $(webAppName)
+              runtimeStack: 'NODE|10.10'
+              package: $(System.ArtifactsDirectory)/drop/$(Build.BuildId).zip
+              startUpCommand: 'npm run start'
--- a/templates/node.js-express-webapp-on-linux.yml
+++ b/templates/node.js-express-webapp-on-linux.yml
@@ -9,7 +9,7 @@ trigger:
 variables:
 
   # Azure Resource Manager connection created during pipeline creation
-  azureSubscription: '{{ azureRmConnection.Name }}'
+  azureSubscription: '{{ azureRmConnection.Id }}'
   
   # Web app name
   webAppName: '{{ webAppName }}'
@@ -22,7 +22,7 @@ stages:
   displayName: Build stage
   jobs:  
   - job: Build
-    displayName: Build job
+    displayName: Build
     pool:
       vmImage: $(vmImageName)
       
@@ -58,7 +58,7 @@ stages:
   condition: succeeded()
   jobs:
   - deployment: Deploy
-    displayName: Deploy job
+    displayName: Deploy
     environment: 'development'
     pool: 
       vmImage: $(vmImageName)    
--- a/templates/node.js-functionapp-on-linux.yml
+++ b/templates/node.js-functionapp-on-linux.yml
@@ -9,7 +9,7 @@ trigger:
 variables:
 
   # Azure Resource Manager connection created during pipeline creation
-  azureSubscription: '{{ azureRmConnection.Name }}'
+  azureSubscription: '{{ azureRmConnection.Id }}'
 
   # Function app name
   functionAppName: '{{ functionAppName }}'
@@ -22,7 +22,7 @@ stages:
   displayName: Build stage
   jobs:  
   - job: Build
-    displayName: Build job
+    displayName: Build
     pool:
       vmImage: $(vmImageName)
       
@@ -59,7 +59,7 @@ stages:
   condition: succeeded()
   jobs:
   - deployment: Deploy
-    displayName: Deploy job
+    displayName: Deploy
     environment: 'development'
     pool: 
       vmImage: $(vmImageName)    
--- a/templates/node.js-functionapp-on-linux.yml
+++ b/templates/node.js-functionapp-on-linux.yml
@@ -58,21 +58,24 @@ stages:
   dependsOn: Build
   condition: succeeded()
   jobs:
-  - job: Deploy
+  - deployment: Deploy
     displayName: Deploy job
-    pool:
-      vmImage: $(vmImageName)
-    
-    steps:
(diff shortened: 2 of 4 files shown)


templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
-  inputs:
-    rootFolderOrFile: '$(System.DefaultWorkingDirectory)'
-    includeRootFolder: false
-    archiveType: zip
-    archiveFile: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
-    replaceExistingArchive: true
+variables:
+
+  # Azure Resource Manager connection created during pipeline creation
+  azureSubscription: '{{ azureRmConnection.Name }}'
```
bansalaseem: should this be Id?

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
+        archiveFile: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
+        replaceExistingArchive: true
+
+    - task: PublishPipelineArtifact@0
+      displayName: 'Publish pipeline artifacts'
+      inputs:
+        targetPath: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
+
+- stage: Deploy
+  displayName: Deploy stage
```
bansalaseem: We should make this a deploy job so that an empty environment gets created and we get traceability

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
+      
+    steps:
+    - task: NodeTool@0
+      inputs:
+        versionSpec: '10.x'
+      displayName: 'Install Node.js'
+
+    - script: |
+        npm install
+        npm run build --if-present
```
vinodkumar3: This will fail if package.json file is not present. For eg in plain Javascript projects

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
+    pool:
+      vmImage: $(vmImageName)
+      
+    steps:
+    - task: NodeTool@0
+      inputs:
+        versionSpec: '10.x'
+      displayName: 'Install Node.js'
+
+    - script: |
```
vinodkumar3: Seems you are not building extensions.csproj file.
I'll recommend to test this out end to end with a newly created function application having explicit extensions like Queue Trigger, this will show error on Function Web App mentioning as "Extensions are not registered"

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
+        npm install
+        npm run build --if-present
+        npm run test --if-present
```
vinodkumar3: You can use npm prune --production to help user to automatically remove extraneous dependencies

templates/node.js-functionapp-on-linux.yml
```diff
@@ -0,0 +1,78 @@
...
+      vmImage: $(vmImageName)
+      
+    steps:
+    - task: NodeTool@0
+      inputs:
+        versionSpec: '10.x'
+      displayName: 'Install Node.js'
+
+    - script: |
+        dotnet build --runtime ubuntu.16.04-x64 --output './bin/'
```
vinodkumar3: Will it fail if there is not any .csproj or .sln file in the repo? Please check!

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
 trigger:
 - master
 
-pool:
-  vmImage: 'ubuntu-latest'
-
```
RoopeshNair: Templates should be multi-stage. CI jobs in build stage and CD jobs in deploy stage

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,74 @@
...
-    rootFolderOrFile: '$(System.DefaultWorkingDirectory)'
-    includeRootFolder: false
-    archiveType: zip
-    archiveFile: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
-    replaceExistingArchive: true
+variables:
+
+  # Azure Resource Manager connection created during pipeline creation
+  azureSubscription: '{{ azureRmConnection.Name }}'
+
```
RoopeshNair: - Move this to deploy stage and 
- add `PublishPipelineArtifact@0` task or `upload` macro in the CI job (whichever is deployed)
- the web app task should be in `deployment` job
- Temporarily, we have to include `DownloadPipelineArtifact@1` task to download the artifact. We will remove this when @kasubram, @niadak PR for auto-injection of download steps, and checkout, download macro support is added to the `deployment` job
- for `environment`: we can use `development` for now, @chshrikh is checking feasibility of supporting variables, if we can pull it through then we can use a variable for the format $(build.repository.name)-dev or something similar
@kmkumaran, @azooinmyluggage, @ashokirla  fyi

RoopeshNair: Above comments apply to all templates

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Name works and is more user-friendly.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: We are recommending node.js templates only if package.json is found.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Had a discussion with @azooinmyluggage and @shashankbarsin. They suggested to keep the template minimal.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Also, this step doesn't fail even thought error is encountered by the script.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Though, name is not rename-safe, but id is. Not sure what should we do.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Done

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Changing the same in this PR. Am I missing something ?

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Updated the PR.

templates/node.js-functionapp-on-linux.yml
ajinkya599: The script logs an error, but the step doesn't fail and the pipeline proceeds. @azooinmyluggage to finalize - whether to keep this in the template or not ?

templates/node.js-express-webapp-on-linux.yml
```diff
@@ -6,35 +6,77 @@
...
+  # Agent VM image name
+  vmImageName: 'ubuntu-latest'
+
+stages:
+- stage: Build
+  displayName: Build stage
+  jobs:  
+  - job: Build
+    displayName: Build job
```
RoopeshNair: We don't need `job` suffix for the displayName. Currently it shows up as 'Build job job'

RoopeshNair: Same for deploy job

templates/node.js-functionapp-on-linux.yml
RoopeshNair: @azooinmyluggage - do we need to use NodeTool@0 task here, just a simple install script one liner will do?

azooinmyluggage: @RoopeshNair just a best practice. Ensure there is a one node version in use unless changed deliberately.

ajinkya599: Proceeding with the check-in. Created an internal tracking item to verify whether we need it or not.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Using the Id as per discussion.

templates/node.js-express-webapp-on-linux.yml
ajinkya599: Done

PR #173, opened 2019-04-16
Modifying the template to consume information from KubernetesResource

rakol-ms:
This PR removes the environment object and introduces a new object of `environmentResource:kubernetes`. This helps in being explicit about the type of objects that are being created


atbagga:
@rakol-ms  kubernetesConnection is not used in the template. It should not be a required parameter in the json. 

cc: @ds-ms @bansalaseem

PR #174, opened 2019-04-16
[K8s] Adding more properties to assets

ds-ms:
This PR adds more properties i.e. `destinationPath` and `description` to the file assets.

### DIFF
--- a/templates/deploy-to-existing-kubernetes-cluster.properties.json
+++ b/templates/deploy-to-existing-kubernetes-cluster.properties.json
@@ -32,11 +32,15 @@
     "assets": [
         {
             "type": "file",
-            "path": "k8s/deployment.yml"
+            "path": "k8s/deployment.yml",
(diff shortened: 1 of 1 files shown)


templates/deploy-to-existing-kubernetes-cluster.properties.json
```diff
@@ -32,11 +32,15 @@
     "assets": [
         {
             "type": "file",
-            "path": "k8s/deployment.yml"
+            "path": "k8s/deployment.yml",
+            "destinationPath": "manifests/deployment.yml",
```
bansalaseem: What will happen if the repo already contains these files?

ds-ms: It would throw an error like this: 
```error
The file manifests/deployment.yml already exists in the branch master of the GitHub repository ds-ms/goSampleApp.
```

PR #176, opened 2019-04-17
Updated name of vm image in node.js-with-react.yml

CannibalKush:
ubuntu-latest is not a valid image identifier, replaced with ubuntu 16.04 as specified [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent)

### DIFF
--- a/templates/node.js-with-react.yml
+++ b/templates/node.js-with-react.yml
@@ -7,7 +7,7 @@ trigger:
 - master
 
 pool:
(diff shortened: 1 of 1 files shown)


vtbassmatt:
`ubuntu-latest` is valid now.

CannibalKush:
Thanks matt!

PR #180, opened 2019-04-22
Fixing manifest extension to YML and removing kubernetesConnection as required parameter

rakol-ms:
`kubernetesConnection` is removed as a required paramter since it is no longer referred  in template


ds-ms:
@rakol-ms You need to remove `kubernetesConnection` object as well from the properties.json file

PR #181, opened 2019-04-23
Build and push to ACR in docker-container template

ds-ms:


templates/docker-container.yml
```diff
@@ -6,14 +6,35 @@
...
+- repo: self
 
 variables:
-  imageName: 'your-container-image-name:$(build.buildId)'
-  dockerfile: '{{ dockerfilePath }}'
-  buildContext: '{{ buildContextPath }}'
+  # Container registry service connection established during pipeline creation - Docker Hub/ACR/...
+  dockerRegistryServiceConnection: '{{ containerRegistryConnection.Id }}'
+  imageRepository: '{{#toAlphaNumericString imageRepository 50}}{{/toAlphaNumericString}}'
+  containerRegistry: '{{ containerRegistryConnection.Authorization.Parameters.loginServer }}'
```
shashankbarsin: What's the configuration panel going to look like for this template?

Are we asking server URL of the container registry as a text input for all container registries? Or for Azure do we show dropdowns for choosing the container registry based on the subscription chosen by the user (similar to AKS template)?

templates/docker-container.yml
```diff
@@ -6,14 +6,35 @@
 trigger:
```
shashankbarsin: Modify the comments in the first three lines of the file to just one line -
"# Build and push an image to Azure Container Registry"

As it stands, right now this template and the configuration panel that is shown next to it only allow for pushing to an existing ACR. So this comment will reflect the same

templates/docker-container.yml
```diff
@@ -6,14 +6,35 @@
...
-pool:
-  vmImage: 'ubuntu-latest'
+resources:
+- repo: self
 
 variables:
-  imageName: 'your-container-image-name:$(build.buildId)'
-  dockerfile: '{{ dockerfilePath }}'
-  buildContext: '{{ buildContextPath }}'
+  # Container registry service connection established during pipeline creation - Docker Hub/ACR/...
```
shashankbarsin: If this template and the accompanying configuration panel just address building and pushing an image to existing ACR, remove Docker Hub and ... from this line

PR #203, opened 2019-05-16
Introducing deployment job properties

RoopeshNair:
Support for continueOnError, timeouts, variables etc and template references in deployment job

### DIFF
--- a/design/deployment.md
+++ b/design/deployment.md
@@ -69,27 +69,27 @@ The following example snippet demonstrates how a pipeline can refer an environme
 ```YAML
 jobs:
 - deployment: DeployWeb
-    displayName: deploy Web App
-    pool:
-      vmImage: 'Ubuntu-16.04'
-    # records deployment against bookings resource - Kubernetes namespace
-    environment: 'smarthotel-dev.bookings'
-    strategy: 
-      runOnce:
-        deploy:
-          steps:
-            # No need to explicitly pass the connection details
-          - task: KubernetesManifest@0
-            displayName: Deploy to Kubernetes cluster
-            inputs:
-              action: deploy
-              namespace: $(k8sNamespace)
-              manifests: |
-                $(System.ArtifactsDirectory)/manifests/*
-              imagePullSecrets: |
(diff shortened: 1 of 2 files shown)


design/deployment.md
```diff
@@ -0,0 +1,98 @@
...
+
+```yaml
+jobs:
+- deployment: string  # name of the deployment job, A-Z, a-z, 0-9, and underscore
+  displayName: string  # friendly name to display in the UI
+  dependsOn: string | [ string ]
+  environment: string # target environment name and optionally a resource-name to record the deployment history; format: <environment-name>.<resource-name>
+  condition: string
+  continueOnError: boolean  # 'true' if future jobs should run even if this job fails; defaults to 'false'
+  pool: pool # see pool schema
```
kmkumaran: When a pool is not specified, regular job assumes it to be hosted. For deployment job, this assumption may be incorrect. I believe we would want to use the Pool (for VM resources) in Environment implicitly instead.

design/deployment.md
```diff
@@ -0,0 +1,98 @@
...
+  timeoutInMinutes: number # how long to run the job before automatically cancelling
+  cancelTimeoutInMinutes: number # how much time to give 'run always even if cancelled tasks' before killing them
+  variables: { string: string } | [ variable | variableReference ] 
+  strategy: 
+    runOnce: # default runOnce strategy, useful for running the steps sequentially once against the environment.
+      deploy:
+        displayName: string # friendly name to display in the UI
```
kmkumaran: Can we have filters (or tags) added?

design/deployment.md
RoopeshNair: Yes, pool is optional in case of VM targets. For other resources we still rely on pool. There can be some optimizations where user can leverage an existing VM agent for remote work loads, if the deployment job is targeting more than one type of resource (VM, web apps)

design/deployment.md
RoopeshNair: Updated environment schema including `tags` filter for the resources in the environments.md file. URL property is yet to be finalized

PR #205, opened 2019-05-20
Design updates for Pipeline Artifacts (May 2019)

willsmythe:
Updates the existing design for working with pipeline artifacts. Highlights:

1. Switches back to using "publish" terminology to avoid introducing a new term and to avoid a naming misalignment between the keyword (`publish`) and its associated task (_Publish Pipeline Artifact_)

2. Clarifies behavior when different combinations of `path` and `artifact` are specified, including cases involving absolute paths

### DIFF
--- a/design/pipeline-artifacts.md
+++ b/design/pipeline-artifacts.md
@@ -1,29 +1,36 @@
 # Pipeline Artifacts YAML shortcut
 
-**Status: Ready for dev design**
+**Status: Under development**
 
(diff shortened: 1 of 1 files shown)


design/pipeline-artifacts.md
```diff
@@ -159,9 +165,10 @@ jobs:
   - script: ls $(Pipeline.Workspace)
     # listing shows one folder, "foo"
 ```
-In this case the artifact contents for each artifact is downloaded to `$(Pipeline.Workspace)/foo/WebTier/` and `$(Pipeline.Workspace)/foo/DBTier/` respectively.
+In this case the artifact contents for get downloaded to `$(Pipeline.Workspace)/foo/WebTier/` and `$(Pipeline.Workspace)/foo/DBTier/` respectively (separate directories under "foo" for each artifact).
```
infin8x: Extra word 'for'?

andy-mishechkin (approved):

andy-mishechkin:
@willsmythe, may you check and resolve comment, please

Thank you

PR #210, opened 2019-05-22
Ashokirla acr update

ashokirla:
Updated spec for ACR container resource.

### DIFF
--- a/design/pipeline-resources.md
+++ b/design/pipeline-resources.md
@@ -238,6 +238,10 @@ resources:          # types: pipelines | repositories | containers | packages
   - container: string # identifier for the container resource   
     connection: string # service connection (Docker registry) to connect to the image registry;
     image: string # container image name, Tag/Digest is optional; defaults to latest image
+    options: string  # arguments to pass to container at startup
+    env: { string: string }  # list of environment variables to add
+    ports: [ string ] # ports to expose on the container
+    volumes: [ string ] # volumes to mount on the containe
 ```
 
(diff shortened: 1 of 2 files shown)


design/pipeline-resources.md
```diff
@@ -228,17 +228,16 @@ other repositories' checkout directory: `$PIPELINES_RESOURCESDIR/<repository-ide
...
 ```yaml
 resources:          # types: pipelines | repositories | containers | packages
   containers:
-  - container: string # identifier for the container resource      
-    type: enum # type of the registry like ACR, Docker etc. 
-    connection: string # service connection to connect to the image registry, defaults to ACR??
+  - container: string # identifier for the container resource   
+    connection: string # service connection (Docker registry) to connect to the image registry;
     image: string # container image name, Tag/Digest is optional; defaults to latest image
-    options: string # arguments to pass to container at startup
```
vtbassmatt: You can't remove `options`, `env`, `ports`, and `volumes`. They're used in container scenarios today. See https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#resources

ashokirla: Thanks for the feedback Matt. Yes, it should support other container scenarios as you mentioned and we will continue supporting that. Here I have focused on container CD scenarios. I will mention about that.

ashokirla: Updated the schema as suggested

PR #221, opened 2019-05-29
List web apps and function apps for nodeJs and Java templates.

ajinkya599:


templates/node.js-express-webapp-to-linux-on-azure.properties.json
```diff
@@ -8,9 +8,19 @@
...
+            "type": "dataSourcePicklist",
             "required": "true",
-            "displayName":"Web App name"
+            "displayName": "Web App name"
+        }
+    ],
+    "dataSourceBindings": [
+        {
+            "dataSourceName": "AzureRMWebAppNamesByAppType",
+            "endpointParameterName": "azureRmConnection",
```
ds-ms: Won't `endpoint` suffice?

ajinkya599: In task inputs, `endpointId` is present, which is the input corresponding to the endpoint (The value of which would be the endpoint id).
Named it `endpointParameterName` for clarity, since in future we may introduce `endpointId` as well.

ds-ms: I suggest we go with `endpoint` and use a resolver to get the id. 
In your case `{{ azureRmConnection }}` maybe?

PR #223, opened 2019-05-31
Spec for Artifacts tasks

infin8x:


design/artifacts-in-pipelines.md
```diff
@@ -23,103 +19,414 @@ Pipelines should support customers invoking a supported (see later section) tool
...
+
+- 0-1 Azure Artifacts feed name/GUID
+- 0+ npm-typed service connection names/GUIDs
+- Agent's proxy information, if available
+
+#### Actions
+
+##### Authentication
+
+The task looks for an `.npmrc` either in the root of the repo or at the user's provided location and sets appropriate [environment variables](https://docs.npmjs.com/misc/config#environment-variables) for each item provided. For example, if a user provids a service connection for npmjs with an authtoken, the task will set an environment variable:
```
jotaylo: nit: provides

design/artifacts-in-pipelines.md
```diff
@@ -23,103 +19,414 @@ Pipelines should support customers invoking a supported (see later section) tool
...
+
+`npm_config_//pkgs.dev.azure.com/codesharing-demo/_packaging/codesharing-demo/npm/registry/:username=codesharing-demo` 
+
+`npm_config_//pkgs.dev.azure.com/codesharing-demo/_packaging/codesharing-demo/npm/registry/:_password=token-here`
+
+`npm_config_//pkgs.dev.azure.com/codesharing-demo/_packaging/codesharing-demo/npm/registry/:email=npm requires email to be set but doesn't use the value`
+
+##### Proxy
+
+If the agent's proxy information is available, the task will set the `HTTPS_PROXY` and/or `HTTP_PROXY` variables as appropriate.
```
jotaylo: How about NO_PROXY? This one is harder, because it's not just a value we can plumb through, so it would be good to document what it will get set to. We have a list of regexes of the hosts that should be bypassed, so at the moment in the tasks the registries found in the npmrc file are run against the regexes, and if there are any matches, those registries are added to NO_PROXY.

design/artifacts-in-pipelines.md
```diff
@@ -23,103 +19,414 @@ Pipelines should support customers invoking a supported (see later section) tool
...
+- 0+ npm-typed service connection names/GUIDs
+- Agent's proxy information, if available
+
+#### Actions
+
+##### Authentication
+
+The task looks for an `.npmrc` either in the root of the repo or at the user's provided location and sets appropriate [environment variables](https://docs.npmjs.com/misc/config#environment-variables) for each item provided. For example, if a user provids a service connection for npmjs with an authtoken, the task will set an environment variable:
+
+`npm_config_//registry.npmjs.org/:_authToken=token`
```
jotaylo: Looks like this was added to npm within the last year, so we should probably have a "min supported version" of npm called out somewhere. Also we should validate that it works with yarn.

design/artifacts-in-pipelines.md
```diff
@@ -23,103 +19,414 @@ Pipelines should support customers invoking a supported (see later section) tool
...
+
+##### Authentication
+
+The task sets appropriate [environment variables](https://github.com/Microsoft/artifacts-credprovider#environment-variables) to allow automatic authentication to all Azure Artifacts feeds within the organization and all external feeds for which service connections are provided, and ensures the credential provider is installed. To help users understand this, we should provide a message in the logs along the lines of "Note: This task sets up authentication via https://github.com/Microsoft/artifacts-credprovider, which requires NuGet 4.9.0+, dotnet 2.1.500+ or 2.2.100+, or msbuild 15.9.0+."
+
+`VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` and `VSS_NUGET_URI_PREFIXES` are the environment variables we will set.
+
+##### Proxy
+
+If the agent's proxy information is available, the task will set the `HTTP_PROXY` variables as appropriate.
```
jotaylo: Same deal with NO_PROXY here.

design/artifacts-in-pipelines.md
infin8x: Added a new section to cover this at the bottom.

infin8x: Re Yarn: [it does](https://yarnpkg.com/lang/en/docs/envvars/#toc-npm-config). Thanks for the reminder to check!

PR #252, opened 2019-07-04
Adding a Docker build only template

rakol-ms:
Signed-off-by: Rajendra Kolli <rakol@microsoft.com>

### DIFF
--- a/templates/docker-build.yml
+++ b/templates/docker-build.yml
@@ -1,0 +1,33 @@
+# Docker
+# Build a Docker image 
+# https://docs.microsoft.com/azure/devops/pipelines/languages/docker
+
+trigger:
+- master
+
+resources:
+- repo: self
+
+variables:
+  dockerfilePath: '{{ dockerfilePath }}'
+  tag: '$(Build.BuildId)'
+  
+  # Agent VM image name
+  vmImageName: 'ubuntu-latest'
+
+stages:
+- stage: Build
+  displayName: Build image stage
+  jobs:  
+  - job: Build
+    displayName: Build
+    pool:
+      vmImage: $(vmImageName)
+    steps:
+    - task: Docker@2
+      displayName: Build an image
+      inputs:
+        command: build
+        dockerfile: $(dockerfilePath)
+        tags: |
+          $(tag)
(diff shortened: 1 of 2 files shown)


templates/docker-build.yml
```diff
@@ -0,0 +1,33 @@
...
+    pool:
+      vmImage: $(vmImageName)
+    steps:
+    - task: Docker@2
+      displayName: Build an image
+      inputs:
+        command: build
+        dockerfile: $(dockerfilePath)
+        tags: |
+          $(tag)
```
ds-ms: I think the default tag itself is $(Build.BuildId); you don't need to add it explicitly maybe?

templates/docker-build.yml
```diff
@@ -0,0 +1,33 @@
...
+  - job: Build
+    displayName: Build
+    pool:
+      vmImage: $(vmImageName)
+    steps:
+    - task: Docker@2
+      displayName: Build an image
+      inputs:
+        command: build
+        dockerfile: $(dockerfilePath)
```
ds-ms: Given that this is being used only once, there's no need for the variable maybe?

templates/docker-build.yml
```diff
@@ -0,0 +1,33 @@
...
+  vmImageName: 'ubuntu-latest'
+
+stages:
+- stage: Build
+  displayName: Build image stage
+  jobs:  
+  - job: Build
+    displayName: Build
+    pool:
+      vmImage: $(vmImageName)
```
ds-ms: Given that it's a single stage, you can hard-code it here?

templates/docker-build.yml
rakol-ms: Done. Removed

templates/docker-build.yml
```diff
@@ -0,0 +1,24 @@
...
+
+trigger:
+- master
+
+resources:
+- repo: self
+
+stages:
+- stage: Build
+  displayName: Build image stage
```
ds-ms: `Build image` should do

templates/docker-build.yml
shashankbarsin: No, let it be added explicitly. It'll be clearer for first time users reading the pipeline definition to understand stuff.

templates/docker-build.yml
shashankbarsin: I agree

PR #253, opened 2019-07-04
Missing input

ljkrause:
input `kubernetesServiceEndpoint` appears to be missing in K8n section


templates/docker-container-to-aks.yml
```diff
@@ -41,6 +41,7 @@ variables:
   ACR.Sku: 'Standard'
   AKS.KubeDeploymentYaml: '$(System.DefaultWorkingDirectory)/KubeDeployment.yml' # Update AKS.KubeDeploymentYaml if you want to use deployment file from repo instead of generated file.
   AKS.DeploymentPort: '$(Container.Port)'
+  AKS.ServiceConnectionId: '$(Azure.ServiceConnectionAKS)'
```
ds-ms: You don't need both the endpoints, right?
If you were to choose this to be a Kubernetes Endpoint type, then you'd have to give extra input to the task, mentioning the connection type.

PR #257, opened 2019-07-11
PHP CD template

chperich:


templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+      runOnce:
+        deploy:
+          steps:
+          - task: AzureWebApp@1
+            displayName: 'Deploy Azure Web App : {{ webAppName }}'
+            inputs:
+              azureSubscription: $(azureSubscription)
+              appName: $(webAppName)
+              package: $(Pipeline.Workspace)/drop/$(Build.BuildId).zip
+              runtimeStack: 'PHP|7.3'
```
vincentdass: we can remove this runtime stack, as it is optional

templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+        sudo update-alternatives --set phpdbg /usr/bin/phpdbg$(phpVersion)
+        sudo update-alternatives --set php-cgi /usr/bin/php-cgi$(phpVersion)
+        sudo update-alternatives --set phar.phar /usr/bin/phar.phar$(phpVersion)
+        php -version
+      workingDirectory: $(rootFolder)
+      displayName: 'Use PHP version $(phpVersion)'
+    
+    - script: composer install --no-interaction --prefer-dist
+      workingDirectory: $(rootFolder)
+      displayName: 'composer install'
```
vincentdass: I assume, composer install wont thow any error, if `composer.json` or `composer.phar` not found.

templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+  environmentName: "$(Build.DefinitionName)-Prod"
+
+  # Root folder under which your composer.json file is available.
+  rootFolder: $(System.DefaultWorkingDirectory)
+
+stages:
+- stage: Build
+  displayName: Build stage
+  variables:
+    phpVersion: 7.2
```
vincentdass: why is this hardcoded?

templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+    - task: ArchiveFiles@2
+      displayName: 'Archive files'
+      inputs:
+        rootFolderOrFile: '$(rootFolder)'
+        includeRootFolder: false
+        archiveType: zip
+        archiveFile: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
+        replaceExistingArchive: true
+    
+    - upload: $(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip
```
vincentdass: nit: can we add a `displayname` here?

templates/php.properties.json
```diff
@@ -1,3 +1,26 @@
...
-  "iconName": "php"
+  "iconName": "php",
+  "parameters": [
+        {
+            "name": "azureRmConnection",
+            "type": "endpoint:azureRm",
+            "required": "true"
+        },
+        {
+            "name": "webAppName",
```
vincentdass: Can we make the dropdown editable?
Few customers may want to add variables here

templates/php.yml
```diff
@@ -1,26 +1,80 @@
 # PHP
-# Test and package your PHP project.
-# Add steps that run tests, save build artifacts, deploy, and more:
+# Build, package and deploy your PHP project.
```
vincentdass: `Build, package and deploy your PHP project to Azure Linux Web App`
Can you please check this with PM?

templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+  # Root folder under which your composer.json file is available.
+  rootFolder: $(System.DefaultWorkingDirectory)
+
+stages:
+- stage: Build
+  displayName: Build stage
+  variables:
+    phpVersion: 7.2
+  jobs:
+  - job:
```
vincentdass: nit: add job name

templates/php.properties.json
```diff
@@ -1,3 +1,26 @@
...
+            "required": "true",
+            "displayName": "Web App name"
+        }
+    ],
+    "dataSourceBindings": [
+        {
+            "dataSourceName": "AzureRMWebAppNamesByAppType",
+            "endpointParameterName": "azureRmConnection",
+            "parameters": {
+                "WebAppKind": "webAppLinux"
```
kmkumaran: Only Linux app support?

templates/php.yml
```diff
@@ -1,26 +1,80 @@
...
+  azureSubscription: '{{ azureRmConnection.Id }}'
+  
+  # Web app name
+  webAppName: '{{ webAppName }}'
+
+  # Agent VM image name
+  vmImageName: 'ubuntu-latest'
+
+  # Environment name
+  environmentName: "$(Build.DefinitionName)-Prod"
```
kmkumaran: Does all characters supported by a Build definition Name supported as valid characters for Environment name? Can you check that?

templates/php.yml
chperich: It does throw an error if composer.json is not present. That's a good point.

I'm not sure if we have to add a if else check though. As, most of the repos would have composer.json. In some flows, we might not find composer.json when your application code is probably under a sub folder. If the task fails, user would know that this needs to be corrected. In that case, he would edit the root folder value. If we add a if else check, we would not be surfacing this very well. Let me check with PM on this one.

templates/php.yml
chperich: > Does all characters supported by a Build definition Name supported as valid characters for Environment name? Can you check that?
I'm not sure if this will be the final format. Usha will check and confirm. It might as well be $(reponame)-Dev.

chperich: I'm not sure if this will be the final format. Usha will check and confirm. It might as well be $(reponame)-Dev.

kmkumaran: Ok. You will still need to ensure all reponame values are valid environment name characters.

templates/php.yml
chperich: We have decided to let the composer install fail. User can remove the task if there is no composer.json.

templates/php.properties.json
chperich: Yes, for now, it is only Linux.

templates/php.yml
chperich: Both Build Definition and repo name allow ".". So, we are using WebApp name as default environment name instead.

kmkumaran: Webapp name need not be unique (like within ASE and outside ASE). You will probably need to handle that case.

templates/php-webapp-to-linux-on-azure.yml
```diff
@@ -0,0 +1,80 @@
+# PHP
+# Build, package and deploy your PHP project to Azure Linux Web App.
+# Add steps that run tests and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/php
```
ds-ms: Does this probably need to be merged with the above line?

templates/php-webapp-to-linux-on-azure.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- stage: Deploy
+  displayName: 'Deploy Web App'
+  dependsOn: Build
+  condition: succeeded()
+  jobs:
+  - deployment: DeploymentJob
+    pool:
+      vmImage: $(vmImageName)
+    environment: $(environmentName)
```
ds-ms: Is this feature rolled out?

templates/php-webapp-to-linux-on-azure.properties.json
```diff
@@ -0,0 +1,26 @@
+{
+  "iconName": "php",
+  "parameters": [
+        {
+            "name": "azureRmConnection",
```
ds-ms: NIT: `azureRMServiceConnection` maybe?

templates/php-webapp-to-linux-on-azure.yml
chperich: Yes.

PR #273, opened 2019-07-29
Python flask CD template

chperich:


templates/python-package-to-windows-webapp-on-azure.yml
```diff
@@ -0,0 +1,101 @@
...
+              WebAppName: $(webAppName)
+              ExtensionsList: python353x86
+              OutputVariable: 'PYTHON_EXT'
+
+          - task: AzureRmWebAppDeployment@4
+            displayName: 'Azure App Service Deploy'
+            inputs:
+              azureSubscription: $(azureServiceConnectionId)
+              WebAppName: $(webAppName)
+              packageForLinux: '$(Pipeline.Workspace)/drop/$(Build.BuildId).zip'
```
vincentdass: can u check this? I think we removed this input

templates/python-package-to-windows-webapp-on-azure.yml
```diff
@@ -0,0 +1,101 @@
...
+        deploy:
+          steps:
+          - task: AzureAppServiceManage@0
+            displayName: 'Install Python Extension'
+            inputs:
+              azureSubscription: $(azureServiceConnectionId)
+              Action: 'Install Extensions'
+              WebAppName: $(webAppName)
+              ExtensionsList: python353x86
+              OutputVariable: 'PYTHON_EXT'
```
vincentdass: this input is disable I gues,, can u check this variable? `ExtensionsList: '$(D)`

templates/python-package-to-windows-webapp-on-azure.yml
```diff
@@ -0,0 +1,101 @@
...
+                )
+                call "$(PYTHON_EXT)/python.exe" -m pip install -r requirements.txt
+                if %errorlevel% NEQ 0 (
+                 echo Failed to install dependencies>&2
+                 EXIT /b 1
+                )
+              WebConfigParameters: '-WSGI_HANDLER "$(appFolderName).app"  -PYTHON_PATH "$(PYTHON_EXT)\python.exe"  -PYTHON_WFASTCGI_PATH "$(PYTHON_EXT)\wfastcgi.py" -STATIC_FOLDER_PATH "$(appFolderName)\static" -appType python_Flask'
+              enableCustomDeployment: true
+              ExcludeFilesFromAppDataFlag: false
+              DeploymentType: webDeploy
```
vincentdass: can we switch to `Run From Package` and observe the behaviour?

templates/python-package-to-windows-webapp-on-azure.yml
```diff
@@ -0,0 +1,101 @@
...
+  azureServiceConnectionId: '{{ azureServiceConnection.Id }}'
+  
+  # Web app name
+  webAppName: '{{ webAppName }}'
+
+  # Agent VM image name
+  vmImageName: 'windows-latest'
+
+  # Environment name
+  environmentName: '{{ webAppName }}'
```
ds-ms: Given that it's only used once, perhaps you can put the `{{ webAppName }}` directly?

templates/python-package-to-windows-webapp-on-azure.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  # Project root folder. Point to the folder containing manage.py file.
+  projectRoot: $(System.DefaultWorkingDirectory)
```
ds-ms: Same here

templates/python-package-to-windows-webapp-on-azure.yml
chandanapericharla: Changed it to package.

templates/python-package-to-windows-webapp-on-azure.yml
chandanapericharla: As discussed offline. Continuing to use it.

templates/python-package-to-windows-webapp-on-azure.yml
chandanapericharla: Changed it to zip deploy.

templates/python-package-to-windows-webapp-on-azure.yml
chandanapericharla: This is just some default value we are trying to populate for the user. We expect this to be edited by the user to suite their needs. Making it a variable will make it easy for them to find it and edit. Would not want to hide it deep in the script.

templates/python-package-to-windows-webapp-on-azure.yml
chandanapericharla: Same here.

chandanapericharla:
> change to zip deploy

Done

templates/python-package-to-windows-webapp-on-azure.yml
ds-ms: That won't be the case right, we would have already created an environment with that name by then?

PR #282, opened 2019-08-02
update nuget to be one task, remove proxy req

elbatk:


elbatk:
Need to redo this. it's not accurate.

PR #286, opened 2019-08-05
Update nuget authenticate spec to match reality

zarenner:


vtbassmatt:
@zarenner is this still relevant and correct?

PR #287, opened 2019-08-06
YAML Template for Python

issacnitin:


templates/python-functionapp-to-linux-on-azure.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - task: UsePythonVersion@0
+      displayName: 'Use Python 3.6'
+      inputs:
+        versionSpec: 3.6
+
+    - bash: |
+        python3.6 -m venv worker_venv
+        source worker_venv/bin/activate
+        pip3.6 install setuptools
```
Hazhzeng: I don't think we need to install the setuptools here. When creating the virtual environment, it should contain the setuptools by default.

templates/python-functionapp-to-linux-on-azure.yml
```diff
@@ -0,0 +1,82 @@
...
+        pip3.6 install -r requirements.txt
+      displayName: 'Install Application Dependencies'  
+      
+    - task: ArchiveFiles@2
+      displayName: 'Archive files'
+      inputs:
+        rootFolderOrFile: '$(System.DefaultWorkingDirectory)'
+        includeRootFolder: false
+        archiveType: zip
```
Hazhzeng: Just a heads-up, we also support squashfs format, that will significantly reduce the cold start time. Since ArchiveFiles task does not support squashfs so far, no action requires.

templates/python-functionapp-to-linux-on-azure.yml
pulkitaggarwl: @Hazhzeng is there something we changed recently ? This was needed when last time we did build designer templates about 4 months back.  
If you can confirm at your end that these are not needed, we will remove them from both the templates and save some build time

Hazhzeng: Yes, the virtual environment created by python3.6 venv already has setuptools inside. It is safe to remove `pip3.6 install setuptools`.

issacnitin: Done, thanks

PR #295, opened 2019-08-08
Fix checkout path name

AArnott:
It *appears* that `root` was inconsistently used here and that `path` was probably what was intended.

Please confirm before completing PR.

If this *was* correct before my change, there lacks any explanation or docs that I can find for what `root` means.


vtbassmatt:
Good catch, we had debated several words but ultimately chose `path`.

PR #308, opened 2019-08-23
Update Gradle.yml to enable test publishing by default

konradpabjan:
Similar to the following PR: https://github.com/microsoft/azure-pipelines-yaml/pull/242

Many customers are not aware of rich test result functionality being available. This will enable test results to show up by default.


davidstaheli:
Feel free to update Ant.yml at the same time.  Maven.yml has this value already set to `true`.  Thanks!

PR #330, opened 2019-09-18
Traceability in unified pipelines with Jira

raiyanalam:


vtbassmatt:
Why does this add a pipeline at the root? This repo doesn't have anything to build or release.

raiyanalam:
You're right Matt. Looks like the later commits were pushed accidentally to the same branch while using Azure Pipelines starter template from UI. Will raise a fresh PR to avoid confusion. Closing this.

PR #332, opened 2019-09-19
Added a macro proposal for review app resources in Environment

azooinmyluggage:
Macro proposal for review app resources in Environment
For more details about the review app feature refer to the spec: https://microsoftapc-my.sharepoint.com/:w:/g/personal/atulmal_microsoft_com/EbSJCOm_ni1Dsa1Fhv5KygkBam3JxXauHlB8GkdUmFR6YA?e=fexk2F

The spec has a sample pipeline yaml which builds and deploys an app to a K8s cluster for review app feature.

### DIFF
--- a/design/environment.md
+++ b/design/environment.md
@@ -119,6 +119,32 @@ If you are targeting the resource by name and don't need to specify tags or reso
 environment: environmentName.resourceName
 ```
 
+## Environment and review app resources (proposed)
+Review app works by deploying every pull request from your Git repository to a dynamic environment resource. Reviewers can see how those changes look as well as work with other dependent services before they’re merged into the main branch and deployed to production.
+
+A new macro will be introduced which get expanded by pipeline to include steps which will 
+1. Create a clone of a resource (dynamically create a new resource based on an existing resource in an environment)
+2. Add the new resource to the environment 
+
+```yaml
+jobs:
+- deployment:
+  environment: 
+     name: smarthotel-dev      
+     resourceName: $(System.PullRequest.PullRequestId) 
+  pool:
+    name: 'ubuntu-latest'
+  strategy:                 
+    runOnce:            
+      pre-deploy: 
+        steps:       
+        - reviewApp: MasterNamespace 
+        # name of the resource in the environment which will be used as a base for creating the new resource for review app.
+        # When reviewApp macro is used, pipeline will expand the macro to include task which will create a clone of that resource and add it to the environment given as input
+        # For K8s the resource maps to a namespace and for App service it will map to a new slot
+        # The subsequent steps can be used to further configure the new resource and deploy changes
+```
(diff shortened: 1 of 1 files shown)


vtbassmatt:
@azooinmyluggage are you going to write up the full description of the task, inputs, etc.?

azooinmyluggage:
> This seems reasonable. Before we add it to the language proper, I'd like to understand the underlying task and its inputs. If the task has a simple enough interface, I'd want to use it directly instead of adding more keywords. And if it's not simple, I'd like to understand how you propose to simplify the interface through a keyword (e.g. if the task needs inputs, what about being a keyword will reduce that burden?).

@vtbassmatt Sorry, I just lost track of this PR. The idea is to make it easy to create a review app environment resource.  When the macro is used pipeline will expand the macro to include K8S task which will create a clone of that resource and add to the environment given as input. Two resource types for which we will enable this first are K8s and Web App.
 
For K8s it will 
     1. Create a new namespace with the name set to resourceName by using the variable specified against resourceName 
     2. Register the new namespace to the Environment as a resource 
     3. Tag the new resource with metadata like: URL, Dynamic Environment identifier, PR-commit details etc. 

For Web app, slot will be created for deploying the pr version of app. Additional steps can be run in the job to configure the new dynamically created resource and then deploy the app 

The task that this will expand to will have the following inputs
- task: reviewApp@0 
            displayName: reviewAppResource 
            name: reviewApp 
            inputs: 
              baseEnvironmentName:  # create reveiw environment resource in this environment
              resourceName: string      # name of an existing resource in the environment which will be used for  resource type information
              reviewResourceName: $(system.pullRequest.sourceBranch

The macro will simplify the yaml and just do the right thing for this scenario.

vtbassmatt:
Cool, thanks. The task would actually look like this, I think:
```yaml
steps:
- task: reviewApp@0
  inputs:
    resourceName: <name>
```

`name` is not required and `displayName` will do the right thing automatically. You can set defaults in `task.json` or even in task code to figure out `baseEnvironmentName` and `reviewResourceName`. I'd really like to start there first -- tasks can be versioned and their entire interface upended, but language features are (more or less) forever.

azooinmyluggage:
> Cool, thanks. The task would actually look like this, I think:
> 
> ```yaml
> steps:
> - task: reviewApp@0
>   inputs:
>     resourceName: <name>
> ```
> 
> `name` is not required and `displayName` will do the right thing automatically. You can set defaults in `task.json` or even in task code to figure out `baseEnvironmentName` and `reviewResourceName`. I'd really like to start there first -- tasks can be versioned and their entire interface upended, but language features are (more or less) forever.

The central idea behind reviewApp is to make it really easy for customers to set up and use. Everything can be either inferred or defaulted, to do the right thing. I think we should start with this simple and minimalistic change and make it work for K8s and Web Apps.

PR #335, opened 2019-09-21
Traceability in Azure Pipelines with Jira

raiyanalam:


design/jira-traceability.md
```diff
@@ -0,0 +1,122 @@
...
+## YAML: 
+
+Since traceability primarily involves interacting with external systems, we propose taking configuration for interaction with these external systems in a separate `integrations` section in YAML. As a beginning, we will support a first class integration with JIRA. In future, we may eventually make this section extensible thereby allowing 3rd party authors to create their own integrations. In this document, we will keep focus on JIRA integration. <em>Details on extensibility design are out of scope for this document.</em>
+
+ 
+
+### Schema:
+
+```yaml
+integrations:
```
vtbassmatt: This should look more like:

```yaml
integrations:
- integration: 'jira'
  connection: string
  #...
```

design/jira-traceability.md
```diff
@@ -0,0 +1,122 @@
...
+
+ 
+
+### Schema:
+
+```yaml
+integrations:       
+  integration: jira # identifier of integration; mandatory;  
+  - connection: string # service connection to JIRA cloud; mandatory;
+    resources: # resource filters. commits only in  these resources will checked for linked JIRA issues. Job status will be posted to JIRA for these issues; optional; defaults to all resources
```
vtbassmatt: We need to resolve the questions around which resource types are supported, which will inform whether this is the right keyword / hierarchy.

design/jira-traceability.md
```diff
@@ -0,0 +1,122 @@
...
+    - resource: string # identifier for the resource;  
+      branches: 
+        include: [ string ] # optional; defaults to all branches.
+        exclude: [ string ] # optional; defaults to none
+    stages: [ string ]  # stage filters. Status of all build/deploy jobs in these stages will be posted to JIRA; optional; Defaults to all stages. 
+    environmentType: [ string ] # Environment type mapping as supported by JIRA; optional; By default all environments map to "unmapped" category;
```
vtbassmatt: This should probably be called "environmentMap" or "environmentSpec". Also, it should be either a direct map of AzDO environment names to JIRA environment types _or_ a map of JIRA environment types to lists of AzDO environment names.

# Option 1
```yaml
# schema
environmentMap: { string: string }

# example
environmentMap:
  stage-US: staging
  East-US: production
  West-US: production
```

# Option 2
```yaml
# schema
environmentMap: { string: [ string ] }

# example
environmentMap:
  staging:
  - stage-US
  production:
  - East-US
  - West-US
```

design/jira-traceability.md
```diff
@@ -0,0 +1,122 @@
...
+![Release1](./images/jira-traceability-deploy-1.png)
+
+Drill down view:
+![Release2](./images/jira-traceability-deploy-2.png)
+
+For the API, the combination of `pipelineId` (pipeline definition identifier), `environmentID` (environment identifier), `deploymentSequenceNumber` (deployment identifier) constitute a unique identifier for deployment. 
+
+
+#### When should the data be pushed from Azure Pipelines:?
+ Since build generation and deployments are done at a job level in azure pipelines, we'll push the status information to Jira at job boundaries. For deployments, azure pipelines already has a separate category of jobs, viz, deployment jobs. We'll invoke deployment APIs at the beginning and end of deployment jobs if all the filter conditions are met. For builds, however, we currently do not have a separate category of jobs. For this, we will assume that any job that publishes an artifact is a build job. We'll invoke build API at the end of these jobs if filter conditions are met.
```
vtbassmatt: As discussed on the call, need to figure out what are the right trigger points and how the UI should reflect in-progress pipelines.

design/jira-traceability.md
vtbassmatt: (As you've currently written it, there can be only one integration since it's a property of the "integrations" object. I think you meant "this is a list of integrations".)

raiyanalam: I have updated the schema to reflect this.

design/jira-traceability.md
raiyanalam: Thanks. I have updated the schema to be more verbose and clearly specify the resource types that are supported

design/jira-traceability.md
raiyanalam: Makes sense. I have changed the schema to take environment details along similar lines.

design/jira-traceability.md
raiyanalam: In order to reach a conclusion, we spoke  to customers to see what they want from scenario perspective. We also took a look at the usage patterns today for environments and resources. Based on this, we believe job boundaries are the right trigger points. We'll discuss this further in a call.

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+## YAML: 
+
+Since traceability primarily involves interacting with external systems, we propose taking configuration for interaction with these external systems in a separate `integrations` section in YAML. As a beginning, we will support a first class integration with JIRA. In future, we may eventually make this section extensible thereby allowing 3rd party authors to create their own integrations. In this document, we will keep focus on JIRA integration. <em>Details on extensibility design are out of scope for this document.</em>
+
+ 
+
+### Schema:
+
+```yaml
+integrations:
```
kmkumaran: If you are modelling this as job property, how are we thinking about extensibility? Would we (job execution) handle integrations with every possible artifact? Instead should we have these as tasks (auto injected in the end) which makes it extensible for anyone to author a task that can report back status to their respective work item system? - Model that is similar to download artifacts that we already have?

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+![Release1](./images/jira-traceability-deploy-1.png)
+
+Drill down view:
+![Release2](./images/jira-traceability-deploy-2.png)
+
+For the API, the combination of `pipelineId` (pipeline definition identifier), `environmentID` (environment identifier), `deploymentSequenceNumber` (deployment identifier) constitute a unique identifier for deployment. 
+
+
+#### When should the data be pushed from Azure Pipelines:?
+ Based on customer interactions, customers prefer seeing the build status as soon as a build completes and deployment status when deployment to an environment is complete. Since build generation and deployments are done at a job level in azure pipelines, we'll push the status information to Jira at job boundaries. For deployments, azure pipelines already has a separate category of jobs, viz, deployment jobs. We'll invoke deployment APIs at the beginning and end of deployment jobs if all the filter conditions are met. For builds, however, we currently do not have a separate category of jobs. For this, we will assume that any job that publishes an artifact is a build job. We'll invoke build API at the end of these jobs if filter conditions are met.
```
madhuv-msft: `We'll invoke deployment APIs at the beginning and end of deployment jobs if all the filter conditions are met.` about this, do we link JIRA with environment/deployment jobs, when the job is passed or any condition? why do we have to call at the start of the deployment? is it to indicate the status like inprogress etc?

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+ 
+
+### Schema:
+
+```yaml
+integrations:       
+- integration: jira # name of integration; mandatory;    
+  connection: string # service connection to JIRA cloud; mandatory;
+  resources: # resource filters. Status will be posted to JIRA for issues that are linked to these resources only; optional; defaults to all resources
+  - repository: string #  identifier of the repository resource as defined in YAML resources section
```
madhuv-msft: when publishing the build job status, should we consider these filters at all?

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+## Interaction with Jira Cloud 
+
+### Pull:
+ For pulling information from JIRA we would just need the service connection of the JIRA server. Using this we can pull details of consumed work items and display the list in  azure pipelines. 
+
+### Push:
+Jira currently supports REST APIs that let external systems push build and deployment details in JIRA. There are 2 separate APIs for [build](https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-builds-0-1-bulk-post) and [deployments](https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-deployments-0-1-bulk-post). While JIRA depends on external services to fill these data, the work item UI is entirely painted by JIRA only. External services do not have any control over the UI.
+
+#### Jira Build UI and APIs:
+The  build results are populated in the development section of the work item UI. On further clicking on it, we see the pipeline runs that were associated with the build.
```
madhuv-msft: should we publish always the status irrespective of the outcome like passed/failed? what about the filters used in integration, should we honor them?

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+1. Service connection of the source repository which contains the commits of the resource. This lets us parse commit messages and obtain the JIRA issues that have been linked to commits. 
+2. Service connection of the Jira server which is used to plan and track the changes done in a resource. This lets us talk to JIRA server and pull/push information.
+3. JIRA specific concept mappings like [Environment Type](https://confluence.atlassian.com/bitbucket/set-up-bitbucket-deployments-968683907.html). Jira uses this  categorization to progressively expose information to users in its UI. Since azure pipeline does not have such a concept, we need to take this mapping input from the user.  
+4. Filters (E.g. branch filters) to restrict the information transfer to JIRA.
+
+
+## Interaction with Jira Cloud 
+
+### Pull:
+ For pulling information from JIRA we would just need the service connection of the JIRA server. Using this we can pull details of consumed work items and display the list in  azure pipelines.
```
madhuv-msft: is this means, whenever we are showing linked workitems from traceability perspective, simply use this JIRA integration and show the data irrespective of the filters mentioned in the integration?

design/jira-traceability.md
```diff
@@ -0,0 +1,150 @@
...
+<br/>
+
+The following configuration will post status to JIRA for images from `zenithworksImageRegistry` container registry that are tagged with `prod-ready`
+
+```yaml
+integrations:       
+- integration: jira 
+  connection: myJiraServiceConnection 
+  resources: 
+  - container: zenithworksImageRegistry
```
madhuv-msft: is this means, if the deployment job has not used this container, we will not publish the data? same with pipeline or with repository?

design/jira-traceability.md
madhuv-msft: `we will assume that any job that publishes an artifact is a build job` is this means, if a job is not publishing anything then we will not push the linking?

raiyanalam: We call deployment APIs whenever the status of deployment job changes. It can be any condition - in progress, succeeded, failed etc. Yes, you're right, we call it at the start to show 'in-progress'

design/jira-traceability.md
raiyanalam: Yes. These filters will be applicable for build job too.

design/jira-traceability.md
raiyanalam: yes, we should publish whenever status has changed. Filters should be honoured. Lets discuss this in detail offline.

design/jira-traceability.md
raiyanalam: yes

design/jira-traceability.md
raiyanalam: Yes. If the deployment job consumes this container, only then we publish data. Similar logic for other resource types as well.

design/jira-traceability.md
raiyanalam: This integration is at a job level, however, future integrations may need to be at a release level/stage level. We would need to evaluate such possibilities before choosing a specific design route like the auto task injection. Since this is an implementation detail, we will take the design discussions separately.

design/jira-traceability.md
raiyanalam: You're right. We'll assume that any non-deployment job that publishes is a build job. We will push information for such build jobs only.

PR #337, opened 2019-09-23
YAML Template for Powershell Function App

issacnitin:


templates/powershell-functionapp-to-windows-on-azure.yml
```diff
@@ -0,0 +1,74 @@
...
+
+- stage: Deploy
+  displayName: Deploy stage
+  dependsOn: Build
+  condition: succeeded()
+
+  jobs:
+  - deployment: Deploy
+    displayName: Deploy
+    environment: 'development'
```
ds-ms: Update it here as well

issacnitin: Done

PR #345, opened 2019-10-04
Spec fixups re: enum

vtbassmatt:


vtbassmatt:
@mjroghelia anything else you want to see fixed?

PR #348, opened 2019-10-09
Multi-checkout spec

vtbassmatt:
Introducing the concept of multi-checkout.


vtbassmatt:
We'll aim to complete this PR by Oct 14, 2019.

design/multi-checkout.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Basic syntax, GitHub source provider
+
+The same syntax works for GitHub:
+
+```yaml
+steps:
+- checkout: self
+- checkout: github://MyGitHubOrg/MyToolsRepo
```
jpricket: Note that here no endpoint is defined, so this must be a public GitHub repo.
We should also verify that it works with public GitHub repos.

design/multi-checkout.md
```diff
@@ -0,0 +1,180 @@
...
+
+The `resources` block already allows a `repositories` list.
+Today, those repositories hold only YAML templates, so we can't blindly check them out automatically.
+That's why an explicit `checkout` step is also required.
+
+However, in multi-checkout scenarios, this leads to unnecessary verbosity.
+First you have to mention the repository up top, give it a name, and then you check it out by name.
+To combat this, we'll add an inline repository syntax.
+Inline syntax will cover the `type`, `name`, and `ref` of a repository resource.
+(If other changes are needed - such as `path` or `endpoint` - users must fall back to explicit resource syntax.)
```
jpricket: `path` is a little ambiguous here since there is a `path` on the checkout task. Is there also one on the resource as well?

vtbassmatt: Oops typo. Will fix.

vtbassmatt:
> I like spec a lot (much better than my design doc :))

Good artists borrow... :wink:

design/multi-checkout.md
```diff
@@ -0,0 +1,181 @@
...
+### Behavior changes in multi-checkout mode
+
+#### Checkout directory
+By default, all repositories will be checked out as if they've been `git clone`d into the `Pipeline.Workspace` directory.
+Git uses the last part of the path, minus any trailing slashes, spaces, and the string `.git`.
+(We have YAML syntax for selecting a different path, and if the user specifies that, we'll use their name.)
+This changes the default directory for the `self` repo: it would be `$(Pipeline.Workspace)/s` if it were the only `checkout` step.
+
+#### System variables
+`Build.SourcesDirectory` points to the directory where the `self` repo was checked out whether in single- or multi-checkout mode.
```
scottboehmer: What will Build.SourcesDirectory be set to if self isn't one of the listed checkout values?

vtbassmatt: Good question. Today if you `checkout: none`, we still create an `s/` directory and make that the `Build.SourcesDirectory`. I think the thing that breaks least is to do the same here. Thoughts @scottboehmer and @jpricket ?

(Someone offline suggested we always make it the directory that contains the YAML pipeline. But that won't work once we have repository triggers, since the repo with the YAML might not be checked out or even available.)

scottboehmer: Same behavior as the `checkout: none` case seems reasonable to me. Is there anything to prevent a user's yaml from claiming the `s/` path for a different repo or are we okay if they do that and end up with `Build.SourcesDirectory` pointing to whichever repo they put at that path?

vtbassmatt: No, and there's also nothing really stopping a user from pointing to `a/`, `b/`, or `testResults/` either. I'm OK with the risk (and of course we'll doc). I should mention all that in the spec.

jpricket: Seems reasonable to me.

design/multi-checkout.md
```diff
@@ -0,0 +1,201 @@
...
+### Non-inline syntax
+
+If you need an endpoint or other extended resource field, you must use explicit resource syntax.
+
+```yaml
+resources:
+  repositories:
+  - repository: tools
+    type: github
+    endpoint: MyServiceConnection
```
shashankbarsin: Will we add trigger support here?

vtbassmatt: I asked Vijay to put triggers in a separate PR, but the short answer is yes :)

design/multi-checkout.md
```diff
@@ -0,0 +1,201 @@
...
+- `refs/tags/<tagname>`
+- commit ID
+
+### Non-inline syntax
+
+If you need an endpoint or other extended resource field, you must use explicit resource syntax.
+
+```yaml
+resources:
+  repositories:
```
shashankbarsin: Instead of having explicit checkout steps inside jobs, another option is to automatically clone all the repositories declared under the repositories keyword. That way, we'll be able to highlight this repositories as first class Git artifacts at the pipeline run details level. Also all agent jobs within that pipeline need not have multiple checkout steps in each of them

vtbassmatt: Two problems I can see:
1. Existing pipelines use repo resources only for templates. It would be a surprising change to have those all suddenly start getting cloned.
2. In a multi-job pipeline, it's likely that you don't need every repo for every job. Thus, you end up needing to say which repos you _don't_ want instead of the ones you do.

simonsabin:
The same repo error. Should the same repo be allowed if it’s different refs. I can see benefit in having a build that validates the upstream branch against the pr branch. What is the reason for raising an error. Is it that the repos wil land in the same folder. Is it therefore a need to error if two repos will be configured with the same path

jpricket:
@simonsabin @vtbassmatt I think this needs clarification. If the user attempts to include a `checkout` task for the same repository resource (a repo mentioned in the resources section), then that should be an error. This makes sense since they would simply be the same sources in the same place. If the user uses the inline syntax or creates multiple resources for the same repo (presumably at different refs), that would work fine because the `checkout` tasks would be referencing different resources.

vtbassmatt:
Good points. Given that we'll let you check out two different refs in the same repo, who's to say that you don't have a good reason for checking out the same ref in two directories? Maybe we should just allow this.

simonsabin:
I’d agree seems like a restriction that would annoy someone without any real reason for the restriction.

As a reason we often have code gen and we expect devs to have run code gen before committing but we have checks in the build to validate they have by running it again.

Simon Sabin

________________________________

design/multi-checkout.md
vtbassmatt: (By the way, I wish we could do this :grin: and if we ever do a "YAMLv2", we will. But for now, we can't.)

vtbassmatt:
This feels pretty good, so I'm merging it. Feedback is still welcomed and we can always open a new PR if major updates are needed.

PR #358, opened 2019-10-16
Propose different behavior for dealing with the `/s` directory

vtbassmatt:
We talked about a different design for where to checkout sources. This PR reflects that discussion.

### DIFF
--- a/design/multi-checkout.md
+++ b/design/multi-checkout.md
@@ -58,26 +58,30 @@ If it's used again, it's OK to re-fetch the same data.
 ### Behavior changes in multi-checkout mode
 
 #### Checkout directory
-By default, all repositories will be checked out as if they've been `git clone`d into the `Pipeline.Workspace` directory.
+By default, all repositories will be checked out as if they've been `git clone`d into the `Build.SourcesDirectory` directory.
 Git uses the last part of the path, minus any trailing slashes, spaces, and the string `.git`.
 (We have YAML syntax for selecting a different path, and if the user specifies that, we'll use their name.)
-This changes the default directory for the `self` repo: it would be `$(Pipeline.Workspace)/s` if it were the only `checkout` step.
+This changes the default directory for the `self` repo: it would be `$(Pipeline.Workspace)/s` if it were the only `checkout` step, but now it will be `$(Pipeline.Workspace)/s/<reponame>`.
 
-The pipeline workspace directory has three well-known subdirectories:
+The pipeline workspace directory has four well-known subdirectories:
+- `s/` for sources
 - `a/` for artifacts
 - `b/` for generated binaries
 - `testResults/` for test results files
 
-We'll recommend not to use these as checkout paths, but we won't attempt to block them.
-Likewise, a customer could choose to checkout a non-`self` repo into the `s` directory, and we won't block that.
+We'll recommend not to use any of these as checkout paths, but we won't attempt to block them.
 
 #### System variables
-`Build.SourcesDirectory` points to the directory where the `self` repo was checked out whether in single- or multi-checkout mode.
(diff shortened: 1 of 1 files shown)


design/multi-checkout.md
```diff
@@ -58,26 +58,30 @@ If it's used again, it's OK to re-fetch the same data.
...
 #### System variables
-`Build.SourcesDirectory` points to the directory where the `self` repo was checked out whether in single- or multi-checkout mode.
-If `self` wasn't checked out, we'll still create an empty `s/` directory and set `Build.SourcesDirectory` to it.
-(This is for back-compat and matches what we do for `checkout: none` today.)
+No change to the contents of system variablies like `Build.SourcesDirectory` or `System.DefaultWorkingDirectory`.
+Those will remain `$(Pipeline.Workspace/s`.
 
-`System.DefaultWorkingDirectory` points to the same place as `Pipeline.Workspace` in multi-checkout mode (which is different than single-checkout mode).
-We believe this preserves the intended semantics of these variables, keeping as many scripts and tasks working as possible.
+**NOTE**: if one or more `checkout`s sets an explicit path outside of `$(Pipeline.Workspace)/s`, it's ambiguous where the "root of source" should be.
```
vtbassmatt: This is the part I really don't like.

design/multi-checkout.md
```diff
@@ -58,26 +58,30 @@ If it's used again, it's OK to re-fetch the same data.
...
+Two decent heuristics occur to me:
+1. If at least one repo is in a subdirectory of `$(Pipeline.Workspace)/s`, use that as the value for both variables.
+2. Unconditionally set both variables to the same value as `Pipeline.Workspace`.
+
+Of the two, the first is less likely to break tasks.
+The second is easier to explain to pipeline authors.
```
jpricket: Another alternative, would be to unconditionally set both variables to `$(Pipeline.Workspace)/s` even though it is empty.

design/multi-checkout.md
jpricket: I agree, but without a way for the yaml to specify the "root of source", we just need to make it easy to reason about.

design/multi-checkout.md
vijayma: #(Pipeline.Workspace)/s unconditionally would be my preference as well. The meaning of the $(Build.SourcesDirectory) is the 's' folder.

vtbassmatt: I think you're both saying that in multicheckout mode we: 
- create the `/s` directory unconditionally
- always set both of those variables to it, regardless of any `path` settings on `checkout` steps

I need to think a bit more about this, but it just might work :grin:

vtbassmatt: This sounds fine.

design/multi-checkout.md
```diff
@@ -58,26 +58,30 @@ If it's used again, it's OK to re-fetch the same data.
...
-The pipeline workspace directory has three well-known subdirectories:
+The pipeline workspace directory has four well-known subdirectories:
+- `s/` for sources
 - `a/` for artifacts
 - `b/` for generated binaries
 - `testResults/` for test results files
 
-We'll recommend not to use these as checkout paths, but we won't attempt to block them.
-Likewise, a customer could choose to checkout a non-`self` repo into the `s` directory, and we won't block that.
+We'll recommend not to use any of these as checkout paths, but we won't attempt to block them.
```
leantk: This statement is confusing to me. We are saying use `s/` to checkout correct?

leantk: Well I'm 6 minutes too late lol

vtbassmatt: Haha, never too late. The net of it is: in multicheckout, we'll always create a `s/` directory and call it `Build.SourcesDirectory`. By default, code will be checked out into directories under it. The user can override our default by saying `path` on the checkout and put their source wherever they want.

PR #371, opened 2019-10-29
Step target restricted mode

vtbassmatt:


vtbassmatt:
We'll leave this open through the end of the week to gather public feedback. Hopefully getting started on this next week, 11/4.

design/step-target-safe-mode.md
```diff
@@ -0,0 +1,87 @@
...
+    image: azcr.io/secureazurecontainer:latest
+
+- script: echo running this step on host
+- script: echo running this step in container
+  target:
+    container: somecontainer
+- script: echo running this step in container, in safe mode
+  target:
+    container: somecontainer
+    commands: safe
```
alex-peck: Assuming we aren't allowing users to specify each individual command, would it be better for this word to be something like `commandSet`?

design/step-target-safe-mode.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `target` property of a step will be expanded to allow either a simple string or a mapping.
+If it's a string, the string is the target of the step (either `host` or a container resource name) and safe mode is off (as if `commands: any` were specified).
+If it's a mapping, the following keys are supported:
+
+```yaml
+- script: ...
+  target:
+    container: string (container name or the word `host`)
+    commands: enum (one of "safe" or "any", with "any" the default)
```
alex-peck: `safe` might infer incorrectly that the alternative is `unsafe`

How about `restricted` instead?

PR #374, opened 2019-11-01
String changes

azooinmyluggage:


templates/deploy-to-existing-kubernetes-cluster.properties.json
```diff
@@ -28,7 +28,7 @@
         {
             "name": "reviewApp",
             "type": "boolean",
-            "displayName": "Enable review app for this pipeline"
+            "displayName": "Enable Review App for this pipeline"
```
ds-ms: I thought we wanted to remove `for this pipeline` from this statement

azooinmyluggage: Done

PR #393, opened 2019-11-19
Update multi-checkout.md

llaagg:
Not sure if it's expected behaviour, but it worked for me that way.


vtbassmatt:
It should check them out into directories as if `git clone <repo>` were passed on the command line. That is, it'll take the final path segment of the URL, truncating ".git" if it exists. Is that not what you see?

llaagg:
now i get it...

PR #396, opened 2019-11-21
Read-only variables

vtbassmatt:


design/readonly-variables.md
```diff
@@ -0,0 +1,58 @@
...
+## Enforcement
+
+For variables marked readonly, the agent should refuse to update them.
+Additionally, if the server is asked to update these variables by the agent, it should also refuse.
+
+## Safe deployment
+
+There's not a lot of support or telemetry for detecting how widely our customers depend on current behavior.
+Therefore, we need to put this behind a feature flag and roll it out safely.
+Ideally this is a server-side feature flag, but if we need an agent-side setting as well, that's acceptable.
```
damccorm: Do we want a sprint where we warn but don't change behavior to give people time to react?

design/readonly-variables.md
```diff
@@ -0,0 +1,58 @@
...
+In fact, we document that this is the case for system-injected pipeline variables, but don't actually enforce it today.
+
+A readonly variable can be set once, then its contents cannot be changed by user code.
+Any attempt to do so generates a warning.
+(This is to help with troubleshooting, in case there are tasks/scripts relying on the older behavior.)
+The warning will be something like:
+> `VariableName` is read-only and can't be changed.
+
+Azure Pipelines itself can still change variables.
+For instance, if the build name is changed using the appropriate logging command, then `Build.BuilderNumber` may be changed to reflect that.
```
damccorm: Just to be clear - the only command that should be affected by this is setVariable, right?

abombss:
I would like that clarification as well. Unfortunately when building on
windows using xplat tasks like bash several variables still have incorrect
path separator  values and need to be rewritten in the tasks before they
can be properly be used, would this effect that or rather just impact
setting variables to be carried forward in the pipeline to downstream tasks?

design/readonly-variables.md
vtbassmatt: Good thought!

design/readonly-variables.md
vtbassmatt: Yes, updating spec.

vtbassmatt:
@abombss thanks for the feedback. Can you clarify something for me: in these tasks, are you expecting to be able to rewrite the pipeline variable, or just updating the local environment variable? The latter would be unaffected - you can party on the environment variables all you want.

abombss:
Rewrite the local would be good enough. No need to setVariable to have it
stateful for tasks downstream.

design/readonly-variables.md
```diff
@@ -0,0 +1,65 @@
...
+> `VariableName` is read-only and can't be changed.
+
+Azure Pipelines itself can still change variables.
+For instance, if the build name is changed using the appropriate logging command, then `Build.BuildeNumber` may be changed to reflect that.
+
+All of the following should be readonly:
+- All system variables (whether from the server or agent)
+- Output variables (those set using `isOutput=true`)
+- Queue-time variables
+- Task- and script-created variables with a new setting, `isReadonly=true`
```
jeschu1: To confirm how exactly would `isReadonly=true` be set?

vtbassmatt: `##vso[task.setVariable variable=myVar;isReadonly=true;]myValue`

design/readonly-variables.md
```diff
@@ -37,9 +37,6 @@ All of the following should be readonly:
 - Queue-time variables
 - Task- and script-created variables with a new setting, `isReadonly=true`
 
-Additionally, agents should no longer overwrite the following important environment variables, even if there's a pipeline variable with the corresponding name:
-- `TODO`
```
damccorm: Why did you end up cutting this? Not super opinionated either way, but I think I'd just like to understand the rationale - I think I see where this could be helpful, but its also definitely a challenging/weird line to draw.

damccorm: FWIW - everything else looks good, was just gating my approval on this

vtbassmatt: Great question. I started trying to enumerate the variables we should "protect" on each platform. I eventually realized two things:
- It's an arms race, albeit a slow one. Apple just moved from `bash` to `zsh`, and there are innumerable shell options on Linux. What environment variables are meaningful to each one, and how do we know when we're in a state to honor that?
- There are narrow legitimate use cases for overriding an environment variable. `GOPATH`/`GOROOT` was my go-to example (though it's less relevant now). I don't feel qualified at this moment to enumerate those use cases and carve out exceptions for them.

Basically, this was always kind of a sideshow to the notion of readonly pipeline variables. Instead of trying to poorly design it in a footnote here, if we care enough, we should open a real feature & design for it.

damccorm: Cool, that makes sense to me!

design/readonly-variables.md
```diff
@@ -0,0 +1,100 @@
...
+> `VariableName` is read-only and can't be changed.
+
+Azure Pipelines itself can still change variables.
+For instance, if the build name is changed using the appropriate logging command, then `Build.BuildNumber` may be changed to reflect that.
+
+All of the following should be readonly:
+- All system variables (whether from the server or agent)
+- Output variables (those set using `isOutput=true`)
+- Queue-time variables
+- Task- and script-created variables with a new setting, `isReadonly=true`
```
damccorm: Nit: Should this be `isReadOnly` - aka fully camel case?

jeschu1: I've seen it as both readonly and ReadOnly in languages.  I'm down with whatever Matt decides, but definitely worth calling out.

vtbassmatt: I suspect it's case-insensitive in the agent anyhow... I don't really have a preference on how we doc it.

damccorm: Oh good point - you're correct that it is. I'm good with whatever in that case

design/readonly-variables.md
```diff
@@ -0,0 +1,100 @@
...
+
+There's not a lot of support or telemetry for detecting how widely our customers depend on current behavior.
+Therefore, we need to put this behind a feature flag and roll it out safely.
+Ideally this is a server-side feature flag, but if we need an agent-side setting as well, that's acceptable.
+
+Plan:
+- Sprint _X_: When the flag is on, warn in situations where a variable would be overwritten.
+Turning off the feature flag turns off this warning (to fix sitautions where a customer is overwhelmed with warnings).
+- Sprint _X+1_: When the flag is on, variables cannot be overwritten.
+Turninfg off the flag turns off the behavior but does _not_ turn off the warning.
```
damccorm: Nit: "Turning"

design/readonly-variables.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Safe deployment
+
+There's not a lot of support or telemetry for detecting how widely our customers depend on current behavior.
+Therefore, we need to put this behind a feature flag and roll it out safely.
+Ideally this is a server-side feature flag, but if we need an agent-side setting as well, that's acceptable.
+
+Plan:
+- Sprint _X_: When the flag is on, warn in situations where a variable would be overwritten.
+Turning off the feature flag turns off this warning (to fix sitautions where a customer is overwhelmed with warnings).
```
damccorm: Nit: "situations"

HughDevlin:
Question: Regarding the vulnerability partially motivating this proposal, that of a spurious single quote appearing at the end of a variable value, reported as a behavior difficult to understand and difficult to resolve by multiple customers - this vulnerability would persist, after the proposal is implemented, for variables without the new readOnly=true parameter? All variables in all Azure YAML scripts currently lack this new parameter. Thank you in advance for your reply.

HughDevlin:
Question: Regarding the aspect of the proposal, to issue a warning message, "Variable-name is read-only and cannot be modified", once this proposal is implemented, would not this warning message be issued under similar conditions to the current undesirable behavior, such as the spurious single quote tacked on the end of variable values? Would not this warning message be just as difficult to explain to customers as the current undesirable behavior? Should we not anticipate the question, how do I suppress the warning? Would we take the position, it's just a warning, ignore it? Do we have a standard regarding issuing messages for which there is no reasonable means of addressing? Thank you in advance for your insights.

damccorm:
> Question: Regarding the vulnerability partially motivating this proposal, that of a spurious single quote appearing at the end of a variable value, reported as a behavior difficult to understand and difficult to resolve by multiple customers - this vulnerability would persist, after the proposal is implemented, for variables without the new readOnly=true parameter? All variables in all Azure YAML scripts currently lack this new parameter. Thank you in advance for your reply.

Yes, behavior won't change for current customers.

The one caveat here, though, is that this isn't a security boundary that we've offered before, so I don't think its quite right to label it a security vulnerability. The attack vector is:

```
step 1 - set variable FOO
step 2 - untrusted task runs which can alter FOO
step 3 - read variable FOO
```

Previously, we would have said that this is an insecure way of running your pipelines, and its certainly not documented as a recommended way of doing things. Now, we're offering a way to do this in a secure way. If we implement this feature and the attack vector still exists when you use read only variables, then I would say its a security vulnerability.

> Question: Regarding the aspect of the proposal, to issue a warning message, "Variable-name is read-only and cannot be modified", once this proposal is implemented, would not this warning message be issued under similar conditions to the current undesirable behavior, such as the spurious single quote tacked on the end of variable values? Would not this warning message be just as difficult to explain to customers as the current undesirable behavior? Should we not anticipate the question, how do I suppress the warning? Would we take the position, it's just a warning, ignore it? Do we have a standard regarding issuing messages for which there is no reasonable means of addressing? Thank you in advance for your insights.

That's a pretty good point. We should probably point to a doc in our warning message explaining common reasons this might happen (e.g. its a system variable, its already been set, bash + `set -x`, etc...), as well as common mitigations. We could also consider offering a variable to suppress these warnings (e.g. this might be useful if you're running bash with `set -x` and don't want to get a warning with every `##` command) and docing that in the help page. @vtbassmatt thoughts?

vtbassmatt:
We [document the interaction with `set -x`](https://docs.microsoft.com/azure/devops/pipelines/troubleshooting#variables-having--single-quote-appended). It's a good point about the warning potentially being irritating, though I think it's strictly better than the current "silently make the variable wrong". I'd like to leave it in for now as we roll this out, and I'll be watching the feedback channels to see if we need to add a feature to suppress it.

HughDevlin:
Question: Regarding the comment above suggesting we document the script commands that may produce the undesirable behavior, and should be avoided in Azure YAML scripts which include invocations of ##vso, or used in conjunction with some kind of mitigation, do we know how many there are? Thank you in advance for your ideas.

design/readonly-variables.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Example
+
+This example YAML shows all the kinds of readonly variables.
+
+```yaml
+variables:
+- name: first
+  value: one
+  readonly: true  # new syntax marking this variable readonly
```
jeschu1: @vtbassmatt , this spec doesn't mention marking a variablegroup as readonly

FWIW that would be very hard to do.

vtbassmatt: Alright, let's wait and see if anyone wants that before we invest.

Jon889:
This will break the ability to cancel a job from a script as described here: https://github.com/Microsoft/azure-pipelines-agent/issues/1743

Really it would be great if there was a proper way to cancel a job so that it shows up as skipped or aborted or something. (Our use case is cancel the job if it's not relevant, where the relevancy is decided based on commit message)

damccorm:
Hey @Jon889 - if you're ok with cancelling the whole pipeline, you should be able to do this by hitting an endpoint - https://docs.microsoft.com/en-us/rest/api/datafactory/pipelineruns/cancel

With that said, this is a bit of a weird pattern; our canceled state generally means "cancelled by user" - this is actually an explicit decision on our end (for example, microsoft/azure-pipelines-tasks#11845) - so I don't see us officially supporting this via a task or an agent primitive.

You can't cancel a single job, because what would that mean for the broader pipeline's status?

The correct way to implement this pattern is to do your check, set a variable (if single-job) or output variable (if the answer affects multiple jobs), and check that variable in the condition for subsequent tasks. Something like:

```yaml
steps:
- script: |
    # compute "should I continue" somehow
    echo ##vso[task.setVariable variable=continue]true
- task: foo@1
  condition: and(succeeded(), eq(variables.continue, 'true'))
- task: bar@1
  condition: and(succeeded(), eq(variables.continue, 'true'))
```

Overwriting system variables was never supported and was always documented as not working. We didn't always enforce that in the context of a single job, which was a mistake on our part. The correct way to change a job's status is to use `##vso[task.complete]` (and if one went through that supported path, it would be obvious that the job can't be canceled - that was correctly blocked).

Note that we had a broader discussion on this specific application in https://github.com/microsoft/azure-pipelines-yaml/issues/439 (most of what I dropped in this comment is just copy/pasted from there), I'd recommend following up in that thread if you have any further thoughts here.

beatcracker:
This will impact me as a customer, is there any way to make it configurable? See [my comment](https://github.com/microsoft/azure-pipelines-yaml/issues/439#issuecomment-597695102) in #439

PR #399, opened 2019-11-26
Agent no-update mode

vtbassmatt:


design/controlling-agent-updates.md
```diff
@@ -0,0 +1,66 @@
...
+
+The agent is considered an extension of the service.
+As such, we have an auto-update mechanism to ensure that new features can rely on the agent support they require.
+Conceptually, it's no different from binaries on the app tier or JavaScript served to the web frontend.
+
+For instance, tasks express a minimum agent version required.
+If a selected agent doesn't meet that requirement, the agent is first asked to update itself.
+Other ways an agent can be updated include using a feature which demands an evergreen agent (YAML pipelines, for instance) or clicking the "upgrade all agents" button in a pool.
+
+For some highly-regulated customers, this kind of evergreen infrastructure is at odds with their compliance needs.
```
edglas1: What are these compliance needs? And what does the customer do before pulling in a new agent so that use of the agent is compliant?

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,66 @@
...
+It must also indicate that the customer opted into this behavior.
+This way, the customer can decide whether to update their agents or edit the pipeline to remove the offending feature.
+For example:
+> This pool doesn't allow agent updates.
+> An administrator can alter this setting on the project pool page.
+> Task `MyTask` version 1.2.3 requires agent version 2.175.1, which could not be satisfied.
+
+Manual agent updates (i.e. clicking the button in the web) should still send update requests as normal.
+
+## Potential problems & solutions
```
jpricket: I think the Pool/Queue UI should make it clear what version each agent is on and ideally flag any agents that have not been updated for any reason.

design/controlling-agent-updates.md
vtbassmatt: Good questions, updating the doc.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,66 @@
...
+For some highly-regulated customers, this kind of evergreen infrastructure is at odds with their compliance needs.
+Such customers today are doing things like ACLing the agent's installation directory so it can't update itself or recompiling the agent without update support.
+This leaves us in a situation where no one is happy:
+we can't service the product correctly, and the customer has to maintain custom configuration / infrastructure to meet their compliance needs.
+
+## Solution
+
+_See below for other solutions considered._
+
+We will add a mode to the project-level pool which puts it into "no update" mode.
```
edglas1: Seems like this needs to be at the Org level pool, right? Or we'd need to run a separate agent per project.

design/controlling-agent-updates.md
vtbassmatt: Agree that they should show the agent version (I thought we already did, but we don't). We could also denote any agents that aren't on the latest-according-to-the-AT version, but that's not necessarily a problem so I wouldn't want the note to be too alarming.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,66 @@
...
+
+The error message must clearly indicate what feature or task demanded agent update.
+It must also indicate that the customer opted into this behavior.
+This way, the customer can decide whether to update their agents or edit the pipeline to remove the offending feature.
+For example:
+> This pool doesn't allow agent updates.
+> An administrator can alter this setting on the project pool page.
+> Task `MyTask` version 1.2.3 requires agent version 2.175.1, which could not be satisfied.
+
+Manual agent updates (i.e. clicking the button in the web) should still send update requests as normal.
```
edglas1: This update should be scoped to the pool. Customers may need to control which version it gets updated to. E.g. customer takes 3 weeks to validate or certify a particular agent, and in the mean time we release a new version. Customer will want to upgrade to the certified version. That starts to get pretty complicated!

design/controlling-agent-updates.md
jpricket: Could we simply have a flag on the agent config that says “AutoUpdate=false”? That attribute could be sent to the server and the server could then evaluate the agents based on this flag. This might satisfy the customers that "don't trust the server". Agents with AutoUpdate turned off will simply reject any calls to update themselves. We could also clearly show this flag in the pool/queue UI.

vtbassmatt: Goooood catch. I don't know why I wrote project-level there 🤦‍♂

design/controlling-agent-updates.md
vtbassmatt: Yep, those updates are already pool-scoped.

design/controlling-agent-updates.md
vtbassmatt: I've added a section in the "rejected approaches" area. But we should talk about this after the holiday, it's possible I'm wrong :grin:

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Potential problems & solutions
+
+**If a small number of agents have been updated, then all jobs will funnel to that agent instead of spreading out across the pool.**
```
edglas1: But if the manual update action, and the setting, are both at the pool level, we shouldn't get into this situation, right?

design/controlling-agent-updates.md
edglas1: @vtbassmatt I think you missed my other comment. Do we have a requirement that the customer can control which version of the agent they update to? See scenario above, where the customer will have validated or certified a specific version of the agent, and we may have shipped new agents in the meantime. Or is the customer responsible getting the new agent bits on the agent machine?

mjroghelia: Given that the primary driver of this feature is people who want to explicitly control agent updates, even with the human overhead that implies, I think the simpler thing is to disallow manual updates altogether when this feature is enabled. That avoids the complexity of evolving the UI to let you dictate which version you want to push down (to @edglas1's point). If this manual update option still exists for a "locked down" pool, we'll also need some way of auditing who has used it, when someone invariably updates a pool outside of their org's normal policy.

edglas1: That makes sense to me. Do we provide any facility for updating the agent? I wonder how Skype does it. They must have 100s of agents.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+For instance, tasks express a minimum agent version required.
+If a selected agent doesn't meet that requirement, the agent is first asked to update itself.
+Other ways an agent can be updated include using a feature which demands an evergreen agent (YAML pipelines, for instance) or clicking the "upgrade all agents" button in a pool.
+
+For some highly-regulated customers, this kind of evergreen infrastructure is at odds with their security/compliance needs.
+- For instance, Customer "K" runs the agent inside a highly change-controlled datacenter.
+Every alteration must be approved by a high-ranking official in their organization after manual source inspection.
+Today, they diff between the release tags of their current version and the version being installed, looking for potential security issues or malicious changes.
+- Customer "N" isn't as concerned about the security aspect but is in a regulated space.
+They must satisfy their auditors that they're aware of all changes taking place inside their datacenter.
```
Joel-K: Would task upgrades fall under this scope as well?  There's no model for that afaik because minor versions of tasks are consumed automatically.  I don't know if this can be fully controlled but at least logged in the build log that inputs included tasks xyz and agent version 123.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+We will add a mode to the organization-level pool which puts it into "no update" mode.
+In this mode, jobs will use the minimum agent version as a demand.
+If no agent can satisfy the demand, the job will fail instead of forcing an agent update.
+
+The error message must clearly indicate what feature or task demanded agent update.
+It must also indicate that the customer opted into this behavior.
+This way, the customer can decide whether to update their agents or edit the pipeline to remove the offending feature.
+For example:
+> This pool doesn't allow agent updates.
+> An administrator can alter this setting on the organization pool page.
```
Joel-K: If this is implemented it would be nice as a pool owner to add a reason for the lock:
```
"Agent upgrades are currently disabled due to the policy imposed by [...], If you would like to read more see <this wiki> or escalate a request to <some email alias>"
```

Personally I would love to just allow all pools to auto upgrade, but in the event of an agent bug, force a rollback to a LKG version and leave a message like:
```
"Agent versions have been pinned to 123.4 due to a bug in 123.5 pending resolution of <this icm>"
```

Right now one of the most difficult aspects of large scale pool management is the in ability to roll back or select a specific version.  If we want upgrade to the latest we can do some basic sanity checks by only upgrading 1/2 agents, but that's not enough to verify all customer scenarios.

If our agents are on 123, a customer wants 124, it would be nice to upgrade only 124, instead of jumping all the way to the latest.  Similarly, if we do upgrade to the latest and there turns out to be issues, we have to rebuild the agent. I'm not aware of a reliable means of downgrading.

mongrelion:
Could you please update the description with a summary of what exactly is the purpose of this PR? :)

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
+# Controlling agent updates
+
+The agent is considered an extension of the service.
+As such, we have an auto-update mechanism to ensure that new features can rely on the agent support they require.
+Conceptually, it's no different from binaries on the app tier or JavaScript served to the web frontend.
+
+For instance, tasks express a minimum agent version required.
+If a selected agent doesn't meet that requirement, the agent is first asked to update itself.
+Other ways an agent can be updated include using a feature which demands an evergreen agent (YAML pipelines, for instance) or clicking the "upgrade all agents" button in a pool.
```
stephenmichaelf: Can RM provide demands from their side to use new features?

design/controlling-agent-updates.md
stephenmichaelf: @Joel-K We have a feature that allows you to disable in the box tasks (the ones that exist when you create an account and are auto updated) as well as marketplace tasks. In this mode only tasks installed into your account through tfx can be used by members of your org. This way you can inventory and control what tasks are installed as well as look at their min agent versions and not update until your Agent can support them. Would that work?

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+
+For some highly-regulated customers, this kind of evergreen infrastructure is at odds with their security/compliance needs.
+- For instance, Customer "K" runs the agent inside a highly change-controlled datacenter.
+Every alteration must be approved by a high-ranking official in their organization after manual source inspection.
+Today, they diff between the release tags of their current version and the version being installed, looking for potential security issues or malicious changes.
+- Customer "N" isn't as concerned about the security aspect but is in a regulated space.
+They must satisfy their auditors that they're aware of all changes taking place inside their datacenter.
+
+Such customers today are doing things like ACLing the agent's installation directory so it can't update itself or recompiling the agent without update support.
```
stephenmichaelf: Maybe add a note that when this happens builds queue up because no Agents can meet their min Agent demands.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+This leaves us in a situation where no one is happy:
+we can't service the product correctly, and the customer has to maintain custom configuration / infrastructure to meet their compliance needs.
+
+## Solution
+
+_See below for other solutions considered._
+
+We will add a mode to the organization-level pool which puts it into "no update" mode.
```
stephenmichaelf: Do we plan to allow the Agent to set this as well for scenarios where the Agent is within a trusted boundary and wants to control its own settings?

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+In this mode, jobs will use the minimum agent version as a demand.
```
stephenmichaelf: Do we plan to also change yaml requesting latest agent always so that we minimize the likelihood of a new agent demand when it may not be needed?

design/controlling-agent-updates.md
Joel-K: If my goal was to control every aspect of a service deployed in my datacenter then the ability to whitelist only certain tasks would be very valuable.  Does this apply to specific versions of tasks or just the tasks themselves?

Personally I don't want to have to verify / approve individual versions.  That sounds like a ton of work.  I just think only pinning agent version is a half measure if you can't also pin tasks to specific minor versions.  If the task lockdown feature accomplishes that then these two features combined should help ensure all changes to the datacenter are performed explicitly.

design/controlling-agent-updates.md
zachariahcox: "can be required to update". 
There is only one way to update. 

"is only supported by the most recent agent version (YAML pipelines for instance)"

design/controlling-agent-updates.md
zachariahcox: Basically, we just do not have an experience to draw something when this process has failed.

design/controlling-agent-updates.md
zachariahcox: "In this mode, jobs will use the minimum agent version as a demand."

This is the current behavior. Should probably read: 

"Azure DevOps always injects the minimum agent version as a demand..."

design/controlling-agent-updates.md
zachariahcox: Regarding the last section about updating to intermediate versions, because rollback is not a action agents can currently take, bugfixes (and effective rollbacks) are always accompanied by a version number increase.

design/controlling-agent-updates.md
```diff
@@ -0,0 +1,85 @@
...
+
+Additionally, the pool UI will be updated to show the agent's version.
+Any agents not fully up to date (as compared to the version on the AT) should be visually marked.
+(Not being 100% up to date isn't necessarily a problem, so it should info-level, not warning-level.)
+
+## Potential problems & solutions
+
+**If a small number of agents have been updated, then all jobs will funnel to that agent instead of spreading out across the pool.**
+This behavior must be well-documented, both in user-facing material and in troubleshooting guides for the team.
+We will inevitably see instances where this setting is on and causing a pool of agents to sit idle while a single agent in the pool is backed up.
```
zachariahcox: It would be cool to have a panel in the pool info where customers could put a custom message and we could draw something like "version upgrade in progress"

design/controlling-agent-updates.md
vtbassmatt: Thanks @Joel-K. The key message here is: we _expect_ the agent to update on our demand. That's the model for 95% (maybe 99%) of our customer base and will remain the default. For customers who - for whatever reason - can't accept that, we need a better failure mode than knocking agents offline permanently. That's what happens today.

If you use a new feature or if a task you depend on suddenly demands a new agent, those pipelines will start failing. Other pipelines should continue working (and might even have less contention for available agents :grin:). I'm arguing that's better than today, where agents start getting picked off and aren't available for any pipeline's use. I have explicit confirmation from at least one of the pseudonymous customers listed here that this is true - they're fast at getting necessary approvals, but not quite fast enough to avoid problems under today's model.

design/controlling-agent-updates.md
vtbassmatt: I would consider that in a separate feature / spec. Doing this at the pool level lets me solve it for both private pools and pool-provider pools. Also, at the agent level, it would boil down to a fairly confusing kind of "capability". I think this would further exacerbate our problems with "why isn't my build running?".

That said, certain customers including "K" are likely to keep recompiling their agent without the update feature until we add an agent-side config. That's not ideal, so I would consider it in addition to this work.

design/controlling-agent-updates.md
vtbassmatt: Yes, good catch.

vtbassmatt: Reworded.

design/controlling-agent-updates.md
vtbassmatt: I do like the idea about having the admin offer up a reason, though! Adding to the spec.

design/controlling-agent-updates.md
vtbassmatt: I'll clarify in the doc. You can get into this state by manually updating a few agents, or clicking the "update this agent" button in the UI (yep, you can do that at the pool level and at the individual agent level).

jbblanchet:
I think one use case that is not documentated here is just about controlling the timing of the updates. It's not about audits and rules, it's mostly about making sure updates to a critical system happen when the risks are minimized (i.e. not on the day a big release is planned).

In such a scenario, some options that would be appreciated would be:
- partial rollout of the update to some agents in the pool, so builds can be tested against that new version without endangering the whole pool. That can be done with manual updates now, but not with automatic.
- ability to manually roll back an update, so if anything breaks, there's a workaround.
- better ability to identify or override version requirements. Why did our builds suddenly start requiring version 2.161.1 yesterday? That is not clear to us, and there was no way of overriding that behavior, or finding what task or step made the upgrade necessary.

By reading your proposal, it seems as if right now, a new version of a task might cause an update in the minimum version of the agent. It would be nice if instead we could control that behavior with the following options:
- always update tasks to latest minor version, even if it bumps minimum agent version (which I believe is the current behavior)
- updates tasks if it's compatible with the current agent version. That would be our favorite setting, as we would get most updates, but that wouldn't force auto-updating of the agents.
- never update tasks automatically (I think this can already be achieved by using the full version number in the task, but I have never tested it)

vtbassmatt:
Live spec review feedback:
- Thought experiment about having the agent be a tiny bootstrapper + task-like payload containing the bulk of it
- In BYOS, since we provide the agent (just like MMS), is this inside our compliance boundary?

vtbassmatt:
Oops, I missed @jbblanchet 's comments - sorry! Thanks for the feedback. This proposal will give you some control towards the scenario you're asking for:
1) We'll stop requesting upgrades quite so often, instead analyzing whether a new feature requires it or not
2) We'll give good error messages in cases where we fail to find an agent of the right version - if it's a task, you can choose to disable or delete the task

We don't want to get into the business of pinning to minor versions of tasks. Same reason as why the service (including its .NET code, the JavaScript it serves up, etc.) stays evergreen: that's how we address bugs, fix security holes, and of course, offer up new features.

vtbassmatt:
We agreed that this spec is in decent shape. While there are a few loose ends to tie up in the design, going to go ahead and land it.

PR #417, opened 2019-12-30
Ashokirla patch 1

ashokirla:
Added specs for package resource support in resources and webhook based triggers.

### DIFF
--- a/design/pipeline-resources.md
+++ b/design/pipeline-resources.md
@@ -115,8 +115,7 @@ resources:
 ```
 
 ### `downloadBuild` for builds
-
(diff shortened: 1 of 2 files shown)


vtbassmatt:
@ashokirla instead of doing the `variables` block, @mjroghelia and I were wondering if you could do something like:

```yaml
resources:       
  webhooks:
  - webhook: MyWebHook
    connection: BookStoreIncomingWebHook
    filters:  
      "$.store.book[0].title": "TrainYourPets"
    parameter: myPayload
```

Then when a customer wants to use the contents of the payload, they can refer to it like this: `${{ parameters.myPayload.store.book[0].Price }}`. The nice thing is that you'll get actual data types rather than strings, so you can do comparisons in the template language. Customers who want to use it in a variable can still do so:

```yaml
# <webhook trigger elided>
variables:
  foo: ${{ parameters.myPayload.store.book[0].Price }}
```

I'm sorry we didn't think of this before!

PR #443, opened 2020-02-05
Update deployment.yml

TonyGorman:
Revised appVersion so it works in K8s 1.16.x


TonyGorman (commented):
added required selector and matchLabels

PR #491, opened 2020-04-29
Use options instead of mavenOptions for Maven task

winniehell:
This avoids overriding the globally defined MAVEN_OPTS which breaks caching for example (https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#maven).

### DIFF
--- a/templates/maven.yml
+++ b/templates/maven.yml
@@ -13,7 +13,7 @@ steps:
 - task: Maven@3
   inputs:
     mavenPomFile: 'pom.xml'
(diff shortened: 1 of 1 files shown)


winniehell:
I'm going to close this because it seems like even with this change, the MAVEN_OPTS from the caching snippet are not recognized. :disappointed:

winniehell:
I created https://github.com/MicrosoftDocs/azure-devops-docs/pull/8142 now to update the documentation.

PR #506, opened 2020-07-16
Improve App Service Linux Deployment

berndverst:
Improves App Service Linux Deployment by directly mounting deployed Zip file into Docker container. Additionally, due to the virtual filesystems symlinks need to be avoided.

These steps here will improve the App Service deployment step speed by 10X.


berndverst:
See projectkudu/kudu#2946 for context

PR #510, opened 2020-08-18
Changed api-version to v1 and added matchLabels to deployment spec

anraghun:
Upgraded the api-version to `apps/v1` as `apps/v1beta1` got deprecated and is causing the pipeline to fail.

### DIFF
--- a/templates/resources/k8s/deployment.yml
+++ b/templates/resources/k8s/deployment.yml
@@ -1,9 +1,12 @@
-apiVersion : apps/v1beta1
+apiVersion : apps/v1
 kind: Deployment
 metadata:
   name: {{#toAlphaNumericString imageRepository 50}}{{/toAlphaNumericString}} 
 spec:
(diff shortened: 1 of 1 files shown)


shigupt202:
NIT: This is supported on kubernetes versions >= 1.16. For previous versions, the existing template is still valid. Should we explicitly call out about the version somewhere?

anraghun:
Will `apps/v1` be backwards compatible for older versions? If not, we should call it out in the doc for "Deploy to AKS" starter pipeline.

shigupt202:
According to the kubernetes [documentation](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/), `apps/v1` is available since `v1.9`. So, it should be compatible with most of the older clusters.

anraghun:
Okay, I'll add the details to the doc as well.

ds-ms (approved):

